PR #3, opened 2020-02-27
🐳(project) use PostgreSQL instead of SQLite for development and tests

madmatah:
## Purpose 

Use PostgreSQL instead of SQLite in Development environment.

## Proposal

- [x] Update docker-compose to start a postgresql container
- [x] Install dockerize
- [x] Update CircleCI configuration
- [x] Don't wait for postgresql container to run linters

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,6 +133,56 @@ jobs:
           name: Lint code with bandit
           command: ~/.local/bin/bandit -qr src/ashley sandbox
 
+  test-back:
+    docker:
+      - image: circleci/python:3.8-buster
+        environment:
+          DJANGO_SETTINGS_MODULE: settings
+          DJANGO_CONFIGURATION: Test
+          DJANGO_SECRET_KEY: ThisIsAnExampleKeyForTestPurposeOnly
+          PYTHONPATH: /home/circleci/fun/sandbox
+          DB_HOST: localhost
+          DB_NAME: ashley
+          DB_USER: fun
+          DB_PASSWORD: pass
+          DB_PORT: 5432
+      # services
+      - image: circleci/postgres:12-ram
+        environment:
+          POSTGRES_DB: test_ashley
+          POSTGRES_USER: fun
+          POSTGRES_PASSWORD: pass
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-back-dependencies-{{ .Revision }}
+      # Attach the frontend production build
+      - attach_workspace:
+          at: ~/fun
+      # While running tests, we need to make the /data directory writable for
+      # the circleci user
+      - run:
+          name: Create writable /data
+          command: |
+            sudo mkdir /data && \
+            sudo chown circleci:circleci /data
+      # Run back-end (Django) test suite
+      #
+      # Nota bene: to run the django test suite, we need to ensure that
+      # Postgresql service is up and ready. To achieve this, we wrap the pytest
+      # command execution with dockerize, a tiny tool installed in the CircleCI
+      # image. In our case, dockerize will wait up to one minute that the database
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -1,6 +1,13 @@
 version: '3.4'
 
 services:
+  postgresql:
+    image: postgres:9.6
```
lunika: Is it possible to use a newer version ? At least version 10 ?

docker-compose.yml
```diff
@@ -13,7 +20,10 @@ services:
       PYLINTHOME: /app/.pylint.d
     env_file:
       - env.d/development/common
+      - env.d/development/postgresql
     ports:
       - "8090:8000"
     volumes:
       - .:/app
+    depends_on:
+      - "${DB_HOST:-postgresql}"
```
lunika: Can you explain why using this environment variable ?

Dockerfile
```diff
@@ -60,13 +60,27 @@ COPY . /app/
 RUN pip uninstall -y ashley
 RUN pip install -e .[dev]
 
+# Install dockerize. It is used to ensure that the database service is accepting
+# connections before trying to access it from the main application.
+ENV DOCKERIZE_VERSION v0.6.1
```
lunika: In lot of projects we choose to remove dockerize from the application image and use directly their image in the `docker-compose` file. Take a look at [marsha's](https://github.com/openfun/marsha/blob/31700e7536417c44fd3ed14354ad21db270467a7/docker-compose.yml) `docker-compose.yml` file

Makefile
```diff
@@ -30,27 +31,27 @@ lint: \
 
 lint-black: ## lint back-end python sources with black
 	@echo 'lint:black started…'
-	@$(COMPOSE_TEST_RUN_APP) black src/ashley sandbox tests
+	@$(COMPOSE_TEST_RUN_APP_NODEPS) black src sandbox tests
```
lunika: If you move dockerize away from the base image, this commit can be removed IMO

.circleci/config.yml
```diff
@@ -133,6 +133,56 @@ jobs:
           name: Lint code with bandit
           command: ~/.local/bin/bandit -qr src/ashley sandbox
 
+  test-back-postgresql:
```
lunika: the suffix `-postgresql` can be removed IMO you are not testing the backend against multiple database engines.

docker-compose.yml
madmatah: Oh, I took it from a project having multiple database containers (mysql + postgresql) and it allows to switch between the two.
Since there is only postgresql here, i'll replace this with `- "postgresql"`.

Makefile
madmatah: In fact, this problem was not related to dockerize but docker-compose (because of the `depends_on = "postgresql"`).

I though it was slower without `--no-depts` option, because before running each linter, docker-compose displayed : `Starting ashley_postgresql_1 ... done`

But in fact, after verification, there is no difference in execution time.
So i'll follow your advice and remove this commit :wink:

madmatah:
Thanks for your feedbacks @lunika !
I addressed all of them.

bin/ashley
```diff
@@ -3,4 +3,6 @@
 declare DOCKER_USER
 DOCKER_USER="$(id -u):$(id -g)"
 
-DOCKER_USER=${DOCKER_USER} docker-compose run --service-ports ashley
+DOCKER_USER=${DOCKER_USER} docker-compose up -d postgresql
```
lunika: I didn't notice this file before. In our project all the run operations are made using a `Makefile`. basically your makefile should have this task `bootstrap`, `build`, `run`, `stop`, `destroy`. But this out of the scope of this PR.

madmatah: Thanks, I take note and i'll address this in another PR.

PR #4, opened 2020-02-27
✨(project) add script to execute management commands

madmatah:
## Purpose

Execution of django management commands in development environment require to execute multiple commands. And, as developers, by definition, we are lazy.

## Proposal

Add a `bin/manage` script allows to execute directly django management commands inside a docker container.

Usage example : `bin/manage createsuperuser`

### DIFF
--- a/bin/ashley
+++ b/bin/ashley
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-declare DOCKER_USER
-DOCKER_USER="$(id -u):$(id -g)"
+# shellcheck source=bin/_config.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
 
-DOCKER_USER=${DOCKER_USER} docker-compose up -d postgresql
-DOCKER_USER=${DOCKER_USER} docker-compose run --rm dockerize dockerize -wait tcp://postgresql:5432 -timeout 60s
-DOCKER_USER=${DOCKER_USER} docker-compose run --rm --service-ports ashley
+_docker_compose up -d postgresql
+_dc_run dockerize dockerize -wait tcp://postgresql:5432 -timeout 60s
+_dc_run --service-ports ashley
--- a/bin/superuser
+++ b/bin/superuser
@@ -3,4 +3,4 @@
 # shellcheck source=bin/_config.sh
 source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
 
-_docker_compose "$@"
+_django_manage createsuperuser "$@"
--- a/env.d/development/common
+++ b/env.d/development/common
@@ -4,5 +4,5 @@ DJANGO_CONFIGURATION=Development
 DJANGO_SECRET_KEY=ThisIsAnExampleKeyForDevPurposeOnly
 
 # Python
-PYTHONUNBUFFERED=1
+PYTHONPATH=/app/sandbox
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,8 @@ ENTRYPOINT [ "/usr/local/bin/entrypoint" ]
 # ---- Development image ----
 FROM core as development
 
+ENV PYTHONUNBUFFERED=1
+
 # Switch back to the root user to install development dependencies
 USER root:root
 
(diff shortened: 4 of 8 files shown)


lunika:
You should maybe take a look at what we did in richie, all scripts are based on this [file](https://github.com/openfun/richie/blob/9dde77c6193650b8987c5d33b804ad0dac2cf0c1/bin/_config.sh) and allow to create multiple small scripts dedicated to `run` or `exec` a command and a container. You can then create one more for management command.
WDYT ?

madmatah:
> You should maybe take a look at what we did in richie, all scripts are based on this [file](https://github.com/openfun/richie/blob/9dde77c6193650b8987c5d33b804ad0dac2cf0c1/bin/_config.sh) and allows to create multiple small scripts dedicated to `run` or `exec` a command and a container. You can then create one more for management command.
> WDYT ?

Thanks, I added a commit to follow your suggestions

bin/ashley
```diff
@@ -1,8 +1,8 @@
...
-DOCKER_USER="$(id -u):$(id -g)"
+# shellcheck source=bin/_config.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
 
-DOCKER_USER=${DOCKER_USER} docker-compose up -d postgresql
-DOCKER_USER=${DOCKER_USER} docker-compose run --rm dockerize dockerize -wait tcp://postgresql:5432 -timeout 60s
-DOCKER_USER=${DOCKER_USER} docker-compose run --rm --service-ports ashley
+_docker_compose up -d postgresql
+_dc_run dockerize dockerize -wait tcp://postgresql:5432 -timeout 60s
+_dc_run --service-ports ashley
```
jmaupetit: Since we define port mapping in the `docker-compose` configuration, is this option required?

env.d/development/common
```diff
@@ -5,4 +5,5 @@ DJANGO_SECRET_KEY=ThisIsAnExampleKeyForDevPurposeOnly
 
 # Python
 PYTHONUNBUFFERED=1
```
jmaupetit: Out of scope here, but should be in the Dockerfile instead.

bin/superuser
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# shellcheck source=bin/_config.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
+
+_django_manage createsuperuser "$@"
```
jmaupetit: Is this shortcut really required and not too much specific as the manage script almost handles it?

bin/ashley
madmatah: Yep, apparently with `docker-compose run`, this option it is required.

`docker-compose status` (with the option):
```
         Name                        Command               State            Ports         
------------------------------------------------------------------------------------------
ashley_ashley_run_1       /usr/local/bin/entrypoint  ...   Up      0.0.0.0:8090->8000/tcp 
ashley_postgresql_1       docker-entrypoint.sh postgres    Up      0.0.0.0:15432->5432/tcp
```

`docker-compose status` (without the option) :
```
         Name                        Command               State            Ports         
------------------------------------------------------------------------------------------
ashley_ashley_run_1       /usr/local/bin/entrypoint  ...   Up                             
ashley_postgresql_1       docker-entrypoint.sh postgres    Up      0.0.0.0:15432->5432/tcp
```

env.d/development/common
madmatah: Ok thanks, I moved it in the `development` image of the Dockerfile.

bin/superuser
madmatah: This shortcut already saved precious seconds of my life :smile: :trollface:

env.d/development/common
jmaupetit: It should also stand for production to ensure unbuffering of stderr/stdout, see: https://docs.python.org/3/using/cmdline.html#cmdoption-u

madmatah: done, thanks ! :+1:

bin/ashley
jmaupetit: Oh I see. The alternative would be to use:

```suggestion
_docker_compose up -d ashley
```

jmaupetit: But it's fine as is.

bin/superuser
jmaupetit: You lazy boy :trollface:

PR #5, opened 2020-02-27
✨(project) add static type checking with mypy

madmatah:
## Purpose

Type hinting using the [PEP 484](https://www.python.org/dev/peps/pep-0484/) notation have been introduced in python 3.5. and I'm convinced that it's a good practice. 

Common bugs can be detected and avoided easily with static type checkers, and [mypy](http://mypy-lang.org/) is one of them.

## Proposal

- [x] Add `mypy` static  type checker in Makefile lint tasks 
- [x] update CircleCI configuration to run `mypy`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -132,6 +132,9 @@ jobs:
       - run:
           name: Lint code with bandit
           command: ~/.local/bin/bandit -qr src/ashley sandbox
+      - run:
+          name: Type-check code with mypy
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -132,6 +132,9 @@ jobs:
       - run:
```
sampaccoud: in your commit message:
s/to to/to
remove a double space

PR #7, opened 2020-02-27
🔧(sandbox) update security settings in sandbox

madmatah:
## Purpose

Ashley is meant to run embedded in iframes on external websites.
With the current configuration, it is not possible because of the following reasons : 

 - django's [clickjacking middleware](https://docs.djangoproject.com/en/2.2/ref/clickjacking/) is sending a [X-Frame-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) header which prevent the `sandbox` project from running inside an iframe on an external website. None of the `X-Frame-Options` directive is suitable if we want to serve ashley as a service for multiple websites at the same time.

- Django CSRF and SESSION cookies are sent with the `SameSite=Lax` flag. If ashley is running inside an iframe on an external website, the cookies sent by ashley are ignored by the user browser.

### Concerning the clickjacking middleware

- Disable the clickjacking middleware

### Concerning the Samesite cookie flag : 

With django 2.2, we can only disable this flag by setting [SESSION_COOKIE_SAMESITE](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-SESSION_COOKIE_SAMESITE) and [CSRF_COOKIE_SAMESITE](https://docs.djangoproject.com/en/2.2/ref/settings/#csrf-cookie-samesite) to `None`.

The result is that the `Samesite` flag will not be sent with the `Set-Cookie` header. We will still have problem with modern browsers that tend to consider that cookies are "Samesite=Lax" by default (see [Chrome announcement](https://www.chromestatus.com/feature/5088147346030592)).

This issue will be [fixed](https://github.com/django/django/commit/b33bfc383935cd26e19a2cf71d066ac6edd1425f) in Django 3.1 and we will be able to send the flag `Samesite=None`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,4 +13,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - lti_provider django application
  - install `django-machina` forum in the `sandbox` project
 
+### Changed
+
+ - Update sandbox settings to be able to run Ashley in an `iframe` on multiple
+   external websites
+
 [unreleased]: https://github.com/openfun/ashley
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -54,6 +54,15 @@ class Base(Configuration):
     # In other cases, you should comment the following line to avoid security issues.
     SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 
+    # Disable Samesite flag in session and csrf cookies, because ashley is meant to
+    # run in an iframe on external websites.
+    # Note : The better solution is to send a flag Samesite=none, because
+    # modern browsers are considering Samesite=Lax by default when the flag is
+    # not specified.
+    # It will be possible to specify CSRF_COOKIE_SAMESITE="none" in Django 3.1
+    CSRF_COOKIE_SAMESITE = None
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,4 +8,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+ - update sandbox settings to be able to run ashley in an iframe on multiple
```
jmaupetit:
```suggestion
 - Update sandbox settings to be able to run Ashley in an `iframe` on multiple
```

sandbox/settings.py
```diff
@@ -43,6 +42,11 @@ class Base(Configuration):
     ALLOWED_HOSTS = []
     SECRET_KEY = values.Value(None)
 
+    # Disable Samesite flag in session and csrf cookies, because ashley is meant to
+    # run in an iframe on external websites.
```
jmaupetit: Please add a note here explaining why this will fail with some browsers and that it will be fixed in Django 3.1 (as you did in your PR description).

madmatah: Good idea ! Done :heavy_check_mark:

PR #8, opened 2020-02-27
✨(app:lti_provider) add lti_provider app

madmatah:
## Purpose

`ashley` will be integrated to a LMS following the LTI specification and will act as a LTI Tool provider.

In order to do that, we should be able to :
 - handle LTI launch requests initiated by the Tool Consumer
 - verify the integrity of the LTI launch requests
 - authenticate user based on LTI launch requests

## Proposal

I introduced the `lti_provider` django application to handle all LTI related operations
It is designed to be a reusable django application without domain specific logic inside.

It is based on [oauthlib](https://github.com/oauthlib/oauthlib).

It provides : 

- A `lti` module to verify LTI launch requests

- A `/launch` url to handle the LTI launch request with configurable behavior on success or failure ( with `LTI_LAUNCH_SUCCESS_HANDLER` and `LTI_LAUNCH_FAILURE_HANDLER` settings)

- A default implementation of a configurable LTI authentication backend (see : `lti_provider.default.backend`)

This app provides a default implementation of the handlers and authentication middleware.
They will be overridden in `ashley` to handle our domain specific needs.


src/fun_lti_provider/default/backend.py
```diff
@@ -0,0 +1,197 @@
...
+            lti_request: the verified LTI request
+
+        Returns:
+            user: the new user to save
+
+        """
+        username = self.get_username_from_lti_request(lti_request)
+        user = USER_MODEL(username=username)
+        user.set_unusable_password()
+        user.email = lti_request.get_param("lis_person_contact_email_primary")
```
jmaupetit:
```suggestion
        user = USER_MODEL(
          username=self.get_username_from_lti_request(lti_request),
          email=lti_request.get_param("lis_person_contact_email_primary")
        )
        user.set_unusable_password()
```

src/fun_lti_provider/launch_params.py
```diff
@@ -0,0 +1,219 @@
...
+    "launch_presentation_width",
+    "launch_presentation_height",
+    "launch_presentation_return_url",
+    "tool_consumer_info_product_family_code",
+    "tool_consumer_info_version",
+    "tool_consumer_instance_guid",
+    "tool_consumer_instance_name",
+    "tool_consumer_instance_description",
+    "tool_consumer_instance_url",
+    "tool_consumer_instance_contact_email",
```
jmaupetit: Can we sort them alphabetically :grimacing: :pray:

src/fun_lti_provider/launch_params.py
```diff
@@ -0,0 +1,219 @@
...
+
+LAUNCH_PARAMS_OAUTH = [
+    "oauth_consumer_key",
+    "oauth_signature_method",
+    "oauth_timestamp",
+    "oauth_nonce",
+    "oauth_version",
+    "oauth_signature",
+    "oauth_callback",
+    "oauth_token",
```
jmaupetit: ditto

src/fun_lti_provider/launch_params.py
```diff
@@ -0,0 +1,219 @@
...
+]
+
+LAUNCH_PARAMS_IS_LIST = [
+    "roles",
+    "role_scope_mentor",
+    "context_type",
+    "accept_media_types",
+    "accept_presentation_document_targets",
```
jmaupetit: same remark for all parameter lists

src/fun_lti_provider/launch_params.py
```diff
@@ -0,0 +1,219 @@
...
+    + LAUNCH_PARAMS_CANVAS
+    + CONTENT_PARAMS_REQUEST
+    + CONTENT_PARAMS_RESPONSE
+    + REGISTRATION_PARAMS
+)
+
+
+class LaunchParams(MutableMapping):
+    """
+    Represents the params for an LTI launch request. Provides dict-like
```
jmaupetit:
```suggestion
    Represents the params for a LTI launch request. Provides dict-like
```

src/fun_lti_provider/launch_params.py
```diff
@@ -0,0 +1,219 @@
...
+        message = "{} is not a valid launch param".format(param)
+        super().__init__(message)
+
+
+class MissingLaunchParamException(LaunchParamException):
+    """Custom Exception thrown when a required LTI parameter is missing in an LTI launch request"""
+
+    def __init__(self, param):
+        message = "missing launch param : {}".format(param)
+        super().__init__(message)
```
jmaupetit: We usually define exceptions in a separated module called... `exceptions.py` :smile:

src/fun_lti_provider/lti.py
```diff
@@ -0,0 +1,196 @@
...
+    """The LTI object abstracts an LTI launch request.
+
+    It provides properties and methods to inspect the launch request.
+    """
+
+    def __init__(self, request):
+        """Initialize the LTI system.
+
+        Parameters
+        ----------
```
jmaupetit: Looks like you are mixing docstring styles. I would prefer sticking to the Google style: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

src/fun_lti_provider/models.py
```diff
@@ -0,0 +1,120 @@
...
+        kwargs: dict
+            Passed onto parent's `save` method
+
+        """
+        self.full_clean()
+        if not self.oauth_consumer_key:
+            self.oauth_consumer_key = "".join(
+                random.choice(OAUTH_CONSUMER_KEY_CHARS)
+                for _ in range(OAUTH_CONSUMER_KEY_SIZE)
+            )
```
jmaupetit: I would suggest to define a separated private method that generates this default key and avoid defining global variables that are only used here. Define them locally instead. Defining a method for this will allow to unit test the random key generation.

src/fun_lti_provider/models.py
```diff
@@ -0,0 +1,120 @@
...
+        if not self.shared_secret:
+            self.shared_secret = "".join(
+                random.choice(SHARED_SECRET_CHARS) for _ in range(SHARED_SECRET_SIZE)
+            )
```
jmaupetit: Same here.

src/fun_lti_provider/settings.py
```diff
@@ -0,0 +1,31 @@
+"""This module contains default settings for fun_lti_provider application"""
+
+from django.conf import settings
+
+# When a LTI launch request is successfully verified, the handler function defined
+# in the LTI_LAUNCH_SUCCESS_HANDLER setting is called with 2 parameters :
+# 1. The request (django.http.request.HttpRequest)
+# 2. The verifier LTI request (fun_lti_provider.lti.LTI)
```
jmaupetit: What about the following?
```suggestion
# 2. The LTI request validator (fun_lti_provider.lti.LTI)
```

src/fun_lti_provider/settings.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# The handler should return a valid HTTP response (django.http.HttpResponse)
+# A default implementation is given in this application as an exemple
```
jmaupetit:
```suggestion
# The handler should return a valid HTTP response (django.http.HttpResponse).
# A default implementation is given in this application as an example.
```

src/fun_lti_provider/settings.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# If you use the default implementation, you must add the following
+# authentication # backend in your AUTHENTICATION_BACKENDS setting :
```
jmaupetit:
```suggestion
# authentication backend in your AUTHENTICATION_BACKENDS setting :
```

src/fun_lti_provider/settings.py
```diff
@@ -0,0 +1,31 @@
...
+#   "fun_lti_provider.default.backend.AuthBackend"
+#
+
+settings.LTI_LAUNCH_SUCCESS_HANDLER = getattr(
+    settings, "LTI_LAUNCH_SUCCESS_HANDLER", "fun_lti_provider.default.handlers.success"
+)
+
+# When a LTI launch request is invalid, the handler function defined in the
+# LTI_LAUNCH_FAILURE_HANDLER is called with the following parameters :
```
jmaupetit:
```suggestion
# LTI_LAUNCH_FAILURE_HANDLER is called with the following parameters:
```

src/fun_lti_provider/settings.py
```diff
@@ -0,0 +1,31 @@
...
+# 1. The request (django.http.request.HttpRequest)
+# 2. The LTIException
+#
+# The handler should return a valid HTTP response (django.http.HttpResponse)
+# A default implementation is given in this application as an exemple
```
jmaupetit:
```suggestion
# The handler should return a valid HTTP response (django.http.HttpResponse).
# A default implementation is given in this application as an example.
```

tests/fun_lti_provider/test_launch_params.py
```diff
@@ -0,0 +1,118 @@
...
+class LaunchParamTestCase(TestCase):
+    """Test the LaunchParam class"""
+
+    def setUp(self):
+        """Override the setUp method to instanciate and serve a request factory."""
+        super().setUp()
+        self.request_factory = RequestFactory()
+
+    def test_only_required_parameters(self):
+        """Test validation of a minimalistic LTI launch request with only required parameters"""
```
jmaupetit:
```suggestion
        """Test validation of a minimalistic LTI launch request with only required parameters."""
```

tests/fun_lti_provider/test_launch_params.py
```diff
@@ -0,0 +1,118 @@
...
+        self._launch_params(
+            {
+                "lti_message_type": "basic-lti-launch-request",
+                "lti_version": "LTI-1p0",
+                "resource_link_id": "df7",
+            }
+        )
+
+    def test_missing_parameters(self):
+        """Test missing required parameters"""
```
jmaupetit: One should add a dot at the end of a docstring first sentence/line.

tests/fun_lti_provider/test_lti.py
```diff
@@ -0,0 +1,220 @@
...
+        signed_parameters = sign_parameters(self._passport, lti_parameters, self._url)
+        lti = self._lti_request(signed_parameters, self._url)
+        lti.verify()
+        return lti
+
+    def _lti_request(self, signed_parameters, url):
+        request = self.request_factory.post(
+            url, data=urlencode(signed_parameters), content_type=CONTENT_TYPE
+        )
+        return LTI(request)
```
jmaupetit: Should be defined before tests for readability.

src/fun_lti_provider/default/backend.py
```diff
@@ -0,0 +1,197 @@
...
+            return None
+
+        if not self.check_lti_request_content(lti_request):
+            return None
+
+        username = self.get_username_from_lti_request(lti_request)
+        logger.debug("User %s authenticated from LTI request", username)
+
+        try:
+            user = USER_MODEL.objects.get(username=username)
```
sampaccoud: should we take care of the `is_active` flag to be able to deactivate a user and forbid access even via LTI?

src/fun_lti_provider/default/backend.py
```diff
@@ -0,0 +1,197 @@
...
+        Returns:
+            bool: True if the LTI request is valid to authenticate the user, or False
+            if the user can't be authenticated.
+
+        Raises:
+            PermissionDenied if the user cannot be authenticated and you dont't want to pass
+            the LTI request to the next configured authentication middleware.
+        """
+        if not lti_request.is_valid:
+            raise PermissionDenied
```
sampaccoud:
```suggestion
            raise PermissionDenied()
```

src/fun_lti_provider/default/backend.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # The LTI request must provide an unique identifier for the user
```
sampaccoud:
```suggestion
        # The LTI request must provide a unique identifier for the user
```

src/fun_lti_provider/lti.py
```diff
@@ -0,0 +1,196 @@
...
+        """LTI roles of the authenticated user.
+
+        Returns
+        -------
+        set
+            normalized LTI roles from the session
+
+        """
+        roles = self.get_param("roles", [])
+        return [*map(str.lower, roles)]
```
sampaccoud: why not?
```suggestion
        return list(map(str.lower, roles))
```

src/fun_lti_provider/models.py
```diff
@@ -0,0 +1,120 @@
...
+    class Meta:
+        """Options for the ``LTIPassport`` model."""
+
+        db_table = "lti_passport"
+        verbose_name = _("LTI passport")
+        verbose_name_plural = _("LTI passports")
+
+    def __str__(self):
+        """Get the string representation of an instance."""
+        return self.oauth_consumer_key
```
sampaccoud: why not return the title which will make more sense to humans?

src/fun_lti_provider/settings.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# If you use the default implementation, you must add the following
+# authentication # backend in your AUTHENTICATION_BACKENDS setting :
+#   "fun_lti_provider.default.backend.AuthBackend"
+#
+
+settings.LTI_LAUNCH_SUCCESS_HANDLER = getattr(
```
sampaccoud: It would be nice to define defaults in your app so that this setting is made optional (see marsha or richie for the pattern).

src/fun_lti_provider/validator.py
```diff
@@ -0,0 +1,154 @@
...
+        Returns:
+            bool: True if the client key is registered and valid
+        """
+        try:
+            passports_count = LTIPassport.objects.filter(
+                oauth_consumer_key=client_key, is_enabled=True
+            ).count()
+            return passports_count > 0
+        except LTIPassport.DoesNotExist:
+            return False
```
sampaccoud: There is a query for this in Django:
```python
return LTIPassport.objects.filter(oauth_consumer_key=client_key, is_enabled=True).exists()
```

src/fun_lti_provider/validator.py
```diff
@@ -0,0 +1,154 @@
...
+            request_token: Request token string, if any.
+            access_token: Access token string, if any.
+
+        Returns:
+            bool: True if the timestamp and nonce has not been used before
+        """
+        # pylint: disable=fixme
+        # TODO: Implement this with a redis server (to have an expiration mechanism) \
+        #  or with a database model, with a cron cleaning older entries.
+        return True
```
sampaccoud: use Django's cache backend?

src/fun_lti_provider/settings.py
madmatah: It's a typo, I meant **verified** LTI request

src/fun_lti_provider/validator.py
madmatah: Thanks for your suggestion !
I'll add an implementation in this PR using django's cache backend.

src/fun_lti_provider/settings.py
madmatah: I don't understand this comment, since the purpose of this module is precisely to provide default values for this application. I used [this pattern](https://stackoverflow.com/a/8429282).

I looked at richie's pattern with `defaults.py` files in apps and it seems to be almost the same.

src/fun_lti_provider/lti.py
madmatah: I hesitated with your suggestion, but I found the syntax with the unpacking operator easier to read.

But I can change :slightly_smiling_face:

src/fun_lti_provider/settings.py
sampaccoud: The difference is that the setting is defined in the defaults.py file from which the code imports it. So it still works if the project has no setting with this name at all.
Said differently, it's like proxy cache for the settings module :wink:

src/fun_lti_provider/lti.py
sampaccoud: Oh then! I think `list` is more pythonic.
I recently read that Python 3.9 introduces a | operator to merge dictionaries and replace the ** notation...
But I realized that js developpers love the star notation and it's fined! :man_shrugging:

src/fun_lti_provider/settings.py
madmatah: We agree on the operating principe :slightly_smiling_face: 

And here, this is the same thing : 
- The [fun_lti_provider/settings.py](https://github.com/openfun/ashley/blob/fun_lti_provider/src/fun_lti_provider/settings.py) (that can be renamed to `defaults.py`) proxifies settings and add a default value if not defined.
- It is imported from the code that use these values (example in [views.py](https://github.com/openfun/ashley/blob/fun_lti_provider/src/fun_lti_provider/views.py#L10))
- The project works if no settings with this name at all is defined in the project : for example, these values are not defined in sandbox's [settings.py](https://github.com/openfun/ashley/blob/fun_lti_provider/sandbox/settings.py) and it works. 

And by writing this, I see that a test is missing for the `view.launch` to prove that it works  as described :slightly_smiling_face:

sampaccoud: ok sorry I just looked at the name and thought it was Django's settings file.

src/fun_lti_provider/default/backend.py
madmatah: You're right! I updated `authenticate()` and `get_user()` to take care of the `is_active` flag.

madmatah:
First, thanks for your feedbacks @jmaupetit and @sampaccoud  ! :pray: 

> 1. I might be confused but it seems that this new application is more a **consumer** than a _provider_ right?

I agree with you : this is confusing!
But according to the [LTI terminology](http://www.imsglobal.org/basic-overview-how-lti-works), Ashley is a Tool Provider and the LMS is a Tool Consumer.

> 2. Avoid using "fun" prefix or suffix in your code as it spreads confusion about our tools' usability outside from the FUN ecosystem.

Concerning the name of the app, i'll rename it to `lti_provider` as suggested by @sampaccoud.

src/fun_lti_provider/default/backend.py
```diff
@@ -0,0 +1,197 @@
...
+from typing import Any, Optional
+
+from django.contrib.auth import get_user_model
+from django.core.exceptions import PermissionDenied
+from django.http import HttpRequest
+
+from fun_lti_provider.lti import LTI
+
+logger = logging.getLogger("fun_lti_provider")
+USER_MODEL = get_user_model()
```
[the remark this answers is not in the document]
madmatah: I tried the CapsWords that you are suggesting and got the following pylint error : 

```
C0103: Constant name "UserModel" doesn't conform to '(([A-Z_][A-Z0-9_]*)|(__.*__)|urlpatterns|logger)$' pattern (invalid-name)
```

So I finished with this upper case version that I don't like either.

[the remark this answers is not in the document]
sampaccoud: yeah that's what I though when I saw jmaupetit's comment. The upper case variable seems legit to me.

src/fun_lti_provider/apps.py
```diff
@@ -0,0 +1,10 @@
+"""Ashley application"""
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class FunLtiProviderAppConfig(AppConfig):
+    """Configuration class for the fun_lti_provider app."""
+
+    verbose_name = _("FUN:LTI Provider")
+    name = "fun_lti_provider"
```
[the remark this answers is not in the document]
madmatah: Thanks for the suggestion, I renamed the app to `lti_provider`

src/fun_lti_provider/models.py
madmatah: I made some modifications after your comment : 
 - The consumer_key and secret key generation are now made inside dedicated methods (`generate_consumer_key` and `generate_shared_secret`)
 - I added tests in [test_models.py](https://github.com/openfun/ashley/blob/fun_lti_provider/tests/fun_lti_provider/test_models.py)
- I replaced the usage of the `random` module with `secrets` to have a stronger random generator and to fix bandit warnings.
- I added entropy in the length of the generated secrets and consumer keys

src/fun_lti_provider/default/backend.py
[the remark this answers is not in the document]
madmatah: Ok thanks, I'll keep the upper case version and mark this conversation resolved.

If you think it's better to rename it to `UsedModel` and remove pylint warning, feel free to unresolve this conversation and add a comment :slightly_smiling_face:

src/fun_lti_provider/models.py
sampaccoud: It would be nice if you can port it to [marsha](https://github.com/openfun/marsha) so we can add bandit there as well.

madmatah: :+1: it's noted on my todo list.
I see that `marsha` is using python >= 3.6 so it won't be a problem to use `secrets`.

src/lti_provider/default/backend.py
```diff
@@ -0,0 +1,206 @@
...
+        username = self.get_username_from_lti_request(lti_request)
+        logger.debug("User %s authenticated from LTI request", username)
+
+        try:
+            user = USER_MODEL.objects.get(username=username)
+        except USER_MODEL.DoesNotExist:
+            user = self.prepare_new_user(lti_request)
+            user.save()
+            logger.debug("User %s created in database", username)
+        if not user:
```
sampaccoud: If I'm not mistaken, this case can never happen.

I'm not very confortable with the way you get the username and email several times in different methods.
Why not get them once in the current method as you do above (and raise if you don't find them so the `check_lti_request_content` is not required anymore...) and pass them to the `prepare_new_user` method instead of passing the lti_request?

src/fun_lti_provider/validator.py
madmatah: Ok, I added an [implementation](https://github.com/openfun/ashley/blob/fun_lti_provider/src/lti_provider/validator.py#L113) with django cache backend with the corresponding [tests](https://github.com/openfun/ashley/blob/fun_lti_provider/tests/lti_provider/test_lti.py#L64-L81)

src/lti_provider/settings.py
```diff
@@ -0,0 +1,38 @@
...
+# A default implementation is given in this application as an example.
+settings.LTI_LAUNCH_FAILURE_HANDLER = getattr(
+    settings, "LTI_LAUNCH_FAILURE_HANDLER", "lti_provider.default.handlers.failure"
+)
+
+# Name of the Django cache to use for the replay attack protection.
+# Timestamp and nonce used to sign LTI request will be stored in this cache
+# to prevent replay attacks.
+settings.LTI_REPLAY_PROTECTION_CACHE = getattr(
+    settings, "LTI_REPLAY_PROTECTION_CACHE", "default"
```
sampaccoud: Then you can just use a cache called "lti_replay" for example and the Django cache framework automatically falls back to default if this cache is not declared in the settings.

src/lti_provider/validator.py
```diff
@@ -0,0 +1,194 @@
...
+        try:
+            cache = caches[cache_name]
+        except InvalidCacheBackendError:
+            logger.warning(
+                "Unable to use cache '%s'. LTI replay protection is DISABLED.",
+                cache_name,
+            )
+            return True
+
+        key = "LTI_TS_NONCE:{}:{}:{}".format(client_key, timestamp, nonce)
```
sampaccoud:
```suggestion
        key = f"LTI_TS_NONCE:{client_key:s}:{timestamp:s}:{nonce:s}"
```

src/lti_provider/validator.py
```diff
@@ -0,0 +1,194 @@
...
+        if cache.get(key) is not None:
+            logger.info(
```
sampaccoud: should it be a warning?

src/lti_provider/validator.py
```diff
@@ -0,0 +1,194 @@
...
+            nonce: The ``oauth_nonce`` request parameter.
+            request: The calling request
+            request_token: Request token string, if any.
+            access_token: Access token string, if any.
+
+        Returns:
+            bool: True if the timestamp and nonce has not been used before
+        """
+
+        cache_timeout = 3600
```
sampaccoud: cache timeout is one of the parameters when declaring a cache backend in Django so maybe you could use it?

src/lti_provider/validator.py
```diff
@@ -0,0 +1,194 @@
...
+            cache = caches[cache_name]
+        except InvalidCacheBackendError:
+            logger.warning(
+                "Unable to use cache '%s'. LTI replay protection is DISABLED.",
+                cache_name,
+            )
+            return True
+
+        key = "LTI_TS_NONCE:{}:{}:{}".format(client_key, timestamp, nonce)
+        if cache.get(key) is not None:
```
sampaccoud: may be a good use case for [getset](https://redis-py.readthedocs.io/en/latest/_modules/redis/client.html#Redis.getset) to avoid race conditions with other processes.

src/lti_provider/default/backend.py
madmatah: > If I'm not mistaken, this case can never happen.

You're right, i'll remove this test.

> I'm not very confortable with the way you get the username and email several times in different methods.
> Why not get them once in the current method as you do above (and raise if you don't find them so the `check_lti_request_content` is not required anymore...) and pass them to the `prepare_new_user` method instead of passing the lti_request?

The goal of this backend is to provide a base implementation that can be extended to add custom domain logic, and that remains compatible with custom user models.
To do that, I had to separate 3 distinct operations : 

- LTI request validation (from a domain point of view) -> `check_lti_request_content`
- User model retrieval -> `get_username_from_lti_request`
- User model creation  -> `prepare_new_user`

For this implementation, your suggestion is totally legit and avoids duplicating username and email retrieval in the LTI request (even if it's not a performance issue, IMO, it's still duplicate code). But it may not fit to every usage ?
 
Here is an example of usage : 

 - I want to only accept request having "role=Instructor" and a non empty `lis_person_sourcedid` field
 - I have a custom user model that use for username a generated value based on the `lis_person_sourcedid` + an information related to the tool consumer that initiated the request.
 - My User model doesn't have any email field (it's possible, right ?)

I think that we can handle most cases with the 3 distinct method I proposed. They get the LTI request as an argument because it contains everything we need in most cases.

Another approach can be to provide a simpler authentication backend and to rewrite it from scratch to add custom logic. But I thought that the proposed solution was more elegant.

madmatah: FYI the current implementation of this backend in ashley looks like this : https://github.com/openfun/ashley/blob/02e9adee4c85b8c6589dc7e934c280d300d7848c/src/ashley/auth/backend.py

src/lti_provider/settings.py
madmatah: good to know, thanks !

madmatah: Hmm it doesn't seem to work when I ask for a cache that is not declared.
I've got an `InvalidCacheBackendError` (_Could not find config for 'lti_replay' in settings.CACHES_) thrown by Django's CacheHandler :
https://github.com/django/django/blob/stable/2.2.x/django/core/cache/__init__.py#L74-L77

src/lti_provider/default/backend.py
sampaccoud: Your right, but username is also not required. Any field can be declared as the unique user identifier as defined here: https://github.com/django/django/blob/master/django/contrib/auth/models.py#L356

I think the best approach is probably to make the code as straightforward as possible for the default user as defined in contrib/auth... and let people write their own backend if they have different needs?

madmatah: I end up with lot of duplicate code when writing the ashley's specific backend, that's what I made this refactoring. But if it helps to better understand the example, i'll write a backend as straigthforward as possible.

sampaccoud: But you can still call the get_username method just that you call it only once? in Django contrib auth they use USERNAME in the generic meaning of "identifier key" so maybe you can as well...

src/lti_provider/settings.py
sampaccoud: oups sorry. So make a try except at the place where you get the cache?

src/lti_provider/validator.py
madmatah: Great I didn't known this syntax

src/lti_provider/settings.py
madmatah: :heavy_check_mark:

src/lti_provider/validator.py
madmatah: Since this is a security related option, I don't feel comfortable with reusing a cache backend option that has been designed for performance. A too small value can cause issues, and a too big value can lead us to accept LTI request signed long time ago.

 I would prefer to make an option in `lti_provider` settings, instead of this hardcoded value, WDYT ?

src/lti_provider/validator.py
madmatah: I've implemented a solution to avoid race conditions, with a single cache operation.

I didn't use `getset` because it is specific to redis and django cache framework does not support it directly (the `get_or_set` method hasn't the same behavior).
I used the `add` operation instead.

src/lti_provider/validator.py
sampaccoud: ok let's leave it like this then.

sampaccoud: Hmm then I think we need a test to confirm that the cache timeout defined in the backend is overriden by what you pass. Otherwise your security issues may be comprised?

src/lti_provider/validator.py
```diff
@@ -0,0 +1,195 @@
...
+
+        try:
+            cache = caches[self.LTI_REPLAY_PROTECTION_CACHE]
+        except InvalidCacheBackendError:
+            logger.debug(
+                "Unable to find cache %s, fallback to %s",
+                self.LTI_REPLAY_PROTECTION_CACHE,
+                DEFAULT_CACHE_ALIAS,
+            )
+            cache = caches[DEFAULT_CACHE_ALIAS]
```
sampaccoud: Django's cache backend can be imported with:
```
from django.core.cache import cache
```

src/lti_provider/validator.py
```diff
@@ -0,0 +1,195 @@
...
+
+        key = f"LTI_TS_NONCE:{client_key:s}:{timestamp:s}:{nonce:s}"
+
+        if not cache.add(key, "1", cache_timeout):
```
sampaccoud: :medal_sports:

src/lti_provider/default/backend.py
madmatah: Here is the [simplified auth backend](https://github.com/openfun/ashley/blob/b6f4650f5d6963e0ddc95cb8d81fab62a23863c7/src/lti_provider/default/backend.py).

src/lti_provider/validator.py
madmatah: As described in [Django's documentation](https://docs.djangoproject.com/en/2.2/topics/cache/#cache-arguments), the `TIMEOUT` option in the backend configuration is a default value, and each [api call](https://docs.djangoproject.com/en/2.2/topics/cache/#basic-usage) allows to set a custom timeout.

Why testing this if it's well documented and this is the responsibility of each cache backend ?

src/lti_provider/validator.py
madmatah: :heavy_check_mark:

src/lti_provider/exceptions.py
```diff
@@ -0,0 +1,39 @@
...
+
+class LaunchParamException(Exception):
+    """Custom Exception related to LTI launch param processing."""
+
+
+class InvalidLaunchParamException(LaunchParamException):
+    """Custom Exception thrown when an invalid parameter is found in a LTI launch request."""
+
+    def __init__(self, param):
+        message = "{} is not a valid launch param".format(param)
```
sampaccoud: replace all format by f strings?

src/lti_provider/default/backend.py
```diff
@@ -0,0 +1,70 @@
...
+            user = USER_MODEL.objects.create_user(username, email=email)
+            logger.debug("User %s created in database", username)
+        if not user.is_active:
+            logger.debug("User %s is not active", user.username)
+            raise PermissionDenied()
+        return user
+
+    @staticmethod
+    def _get_mandatory_param(lti_request: LTI, param: str) -> Any:
+        value = lti_request.get_param(param)
```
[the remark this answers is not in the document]
madmatah: Yes, we have this in the `.pylintrc` : 
```
no-docstring-rgx=^_
```

PR #10, opened 2020-03-04
✨(ashley) add custom authentication backend and LTI handlers for ashley

madmatah:
## Purpose

We should be able to authenticate users from LTI requests signed by the Tool Consumers that we trust.

These users may come from multiple Tool Consumers.

## Proposal

- [x] Custom success/failure handler for ashley
- [x] Custom authentication backend
- [x] Custom user model
- [x] Update settings
- [x] Tests

The permission/group management features will be added later.


src/ashley/models.py
```diff
@@ -0,0 +1,51 @@
+"""Declare the models related to ashley ."""
+
+from django.conf import settings
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from lti_provider.models import LTIConsumer
+
+
+class UserInfo(models.Model):
```
sampaccoud: Have you thought of declaring a custom User with added fields? It's the common practice if what you want is to add the remote_identifier (the username field is already on Django's user).

madmatah: By «_custom User with added fields_», you mean a custom User model ?

Like I said in the docstring of this class, I did this model to avoid this. Because according to the [official documentation](https://docs.djangoproject.com/en/3.0/topics/auth/customizing/#reusable-apps-and-auth-user-model), it is not recommended to extend to implement a custom User model for reusable apps.

The big question is : should we consider `ashley` as a **reusable app**, or a **project** ?
For the moment, I try do design it as a reusable app. WDYT ?

sampaccoud: But it could be an abstract model that people can override... so it is even better than your model because they can add their own fields and/or change the fields you added? I think I saw that django machina is doing it for all their models.

madmatah: For all models it's a good idea, except for the `User` model :
_«A project may use many apps, and two reusable apps that implemented a custom user model couldn’t be used togethe»_

CHANGELOG.md
```diff
@@ -12,5 +12,6 @@ Versioning](https://semver.org/spec/v2.0.0.html).
```
jmaupetit: In your commit message title a verb is missing and the scope is kind of duplicated, I would suggest:

> ✨(ashley) add custom user model, authentication backend and LTI handlers

src/ashley/auth/backend.py
```diff
@@ -0,0 +1,68 @@
...
+from ashley.models import AbstractAshleyUser
+from lti_provider.default.backend import LTIBackend as BaseLTIBackend
+from lti_provider.lti import LTI
+
+logger = logging.getLogger("ashley")
+
+
+class LTIBackend(BaseLTIBackend):
+    """
+    Authentication backend using LTI launch request to authenticate an user.
```
jmaupetit:
```suggestion
    Authentication backend using LTI launch request to authenticate a user.
```

madmatah:
Thanks @jmaupetit, I adressed your feedbacks.

src/ashley/models.py
```diff
@@ -0,0 +1,65 @@
...
+
+from django.contrib.auth.models import AbstractUser
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from lti_provider.models import LTIConsumer
+
+
+class AbstractAshleyUser(AbstractUser):
+    """Abstract user model for ashley.
```
jmaupetit:
```suggestion
    """Abstract user model for Ashley.
```

src/ashley/models.py
```diff
@@ -0,0 +1,65 @@
+"""Declare the models related to ashley ."""
```
jmaupetit:
```suggestion
"""Declare the models related to Ashley."""
```

tests/ashley/test_auth_backend.py
```diff
@@ -0,0 +1,133 @@
...
+        known_user = get_user_model().objects.create_user(
+            "test_auth_backend_user1",
+            email="ashley@example.com",
+            public_username="ashley",
+            lti_consumer=self._consumer1,
+            lti_remote_user_id="ashley",
+        )
+
+        auth_user = self._authenticate()
+        self.assertEqual(known_user, auth_user)
```
jmaupetit: Can you also assert that we have not created a new user by counting the number of users before and after the authentication?

tests/ashley/test_auth_backend.py
```diff
@@ -0,0 +1,133 @@
...
+
+        known_user = get_user_model().objects.create_user(
+            "test_auth_backend_user1",
+            email="ashley@example.com",
+            public_username="ashley",
+            lti_consumer=self._consumer1,
+            lti_remote_user_id="ashley",
+        )
+
+        auth_user = self._authenticate()
```
jmaupetit: Even if those are defaults of your method, I think it's way more explicit to specify some arguments:

```suggestion
        auth_user = self._authenticate("ashley@example.com", "ashley")
```

tests/ashley/test_auth_backend.py
```diff
@@ -0,0 +1,133 @@
...
+        self.assertEqual(known_user, auth_user)
+
+    def test_new_user(self):
+        """Test authentication of a new user."""
+
+        new_user = self._authenticate("newuser@example.com", "new_user")
+        self.assertEqual("new_user", new_user.public_username)
+        self.assertEqual(self._consumer1, new_user.lti_consumer)
+        self.assertEqual("newuser@example.com", new_user.email)
+        self.assertEqual("new_user@test_auth_backend_consumer1", new_user.username)
```
jmaupetit: Also assert that we have incremented the number of total users by one.

tests/ashley/test_auth_backend.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def test_multiple_consumer(self):
```
jmaupetit:
```suggestion
    def test_multiple_consumers(self):
```

tests/ashley/test_auth_backend.py
madmatah: I've rewritten the tests to call the `_authenticate()` helper function with an exhaustive list of LTI parameters.

src/ashley/auth/backend.py
```diff
@@ -0,0 +1,68 @@
+"""Authentication backend for Ashley"""
+
+import logging
+
+from django.contrib.auth import get_user_model
+from django.core.exceptions import PermissionDenied
+
+from ashley.models import AbstractAshleyUser
```
sampaccoud: Maybe go one step further by using [get_user_model](https://docs.djangoproject.com/en/3.0/topics/auth/customizing/#django.contrib.auth.get_user_model) everywhere you use the user and move the non abstract user model declaration to your sandbox so that the user in the Ashley application can really be overriden.

madmatah: This import is only used for type hinting in this file (and maybe it's useless here ?).
I use `get_user_model()` everywhere in the code.

I'll move the non-abstract user model in the sandbox.

src/ashley/admin.py
```diff
@@ -0,0 +1,9 @@
+"""Admin of the Ashley application."""
+from django.contrib import admin
+
+from .models import AshleyUser
```
sampaccoud: @madmatah you sure :stuck_out_tongue:?

madmatah: haha good catch, I just saw that while moving AshleyUser in the sandbox :smile:

src/ashley/auth/backend.py
madmatah: Ok I moved the concrete user implementation in the sandbox.
I had to create an app (`ashley_user`) to be able to set `AUTH_USER_MODEL` in the settings.

sandbox/ashley_user/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""ashley_user is a django application that defines a custom user model implementation."""
+
+# pylint: disable=invalid-name
+default_app_config = "ashley_user.apps.AshleyUserConfig"
```
sampaccoud: maybe ashley_user is not a good name. You can see it as the app in which we customize ashley for the project. I think the common practice in Django is to call it sandbox as well. @jmaupetit WDYT?

madmatah:
I just submitted a new proposal for the custom user model.

I use django-machina's model factory to provide a default implementation of `AbstractUser`.
And it allows anyone to override it with its own user model.

No need to create a new app in the sandbox then.

PR #11, opened 2020-03-09
🐳(project) add docker related rules in the Makefile

madmatah:
## Purpose

`ashley` should have the same basic set of Makefile rules than our other projects like `richie` or `marsha`.

## Proposal

Add the following missing rules to the Makefile : 
- [x] bootstrap
- [x] down
- [x] run
- [x] stop
- [x] migrate

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,33 @@
+# /!\ /!\ /!\ /!\ /!\ /!\ /!\ DISCLAIMER /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\
+#
+# This Makefile is only meant to be used for DEVELOPMENT purpose as we are
+# changing the user id that will run in the container.
+#
+# PLEASE DO NOT USE IT FOR YOUR CI/PRODUCTION/WHATEVER...
+#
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,23 +1,88 @@
...
+	build \
+	migrate
+.PHONY: bootstrap
+
 # -- Docker/compose
 build: ## build the app container
 	@$(COMPOSE) build ashley
 .PHONY: build
 
+down: ## Stop and remove containers, networks, images, and volumes
```
jmaupetit:
```suggestion
down: ## stop and remove containers, networks, images, and volumes
```

Makefile
```diff
@@ -44,33 +114,27 @@ lint-isort: ## automatically re-arrange python imports in back-end code base
...
+lint-pylint: ## lint back-end python sources with pylint
+	@echo 'lint:pylint started…'
+	@$(COMPOSE_TEST_RUN_APP) pylint src sandbox tests
+.PHONY: lint-pylint
 
 test: ## run back-end tests
 	bin/pytest
 .PHONY: test
 
+migrate:  ## Run django migration for the marsha project.
```
jmaupetit:
```suggestion
migrate:  ## run django migration for the marsha project.
```

PR #24, opened 2020-03-18
🐛(sandbox) move factory_boy from dev to sandbox dependencies

madmatah:
## Purpose

The factory-boy module is listed in the dev dependencies, but the sandbox needs it.

## Proposal

- [x] Fix dependencies in `setup.cfg`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,18 +41,18 @@ dev =
     isort==4.3.21
     mypy==0.761
     pyfakefs==3.7.1
-    pylint==2.4.4
     pylint-django==2.0.13
-    pytest==5.3.5
+    pylint==2.4.4
     pytest-cov==2.8.1
     pytest-django==3.8.0
-    factory_boy==2.12.0
+    pytest==5.3.5
 ci =
     twine==2.0.0
 sandbox =
     django-configurations==2.2
-    psycopg2-binary==2.8.4
     dockerflow==2019.10.0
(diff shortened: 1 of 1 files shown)


sampaccoud:
@madmatah why does the sandbox need it? demo fixtures?

madmatah:
> @madmatah why does the sandbox need it? demo fixtures?

There is a `dev_tools` app that is using factories to create a default LTI consumer / passport.
This app is only installed in dev environment, but the [urls.py](https://github.com/openfun/ashley/blob/master/sandbox/urls.py#L11) is importing modules from this app and generates the following error : 

```

File "/app/sandbox/urls.py", line 11, in <module>
  | from dev_tools import urls as dev_urls
  | File "/app/sandbox/dev_tools/urls.py", line 6, in <module>
  | from . import views
  | File "/app/sandbox/dev_tools/views.py", line 12, in <module>
  | from lti_provider.factories import LTIConsumerFactory, LTIPassportFactory
  | File "/usr/local/lib/python3.8/site-packages/lti_provider/factories.py", line 2, in <module>
  | import factory
  | ModuleNotFoundError: No module named 'factory'

```

Another way of fixing it could be with a conditional import in the sandbox views ? WDYT @sampaccoud  ?

sampaccoud:
@madmatah ok. I was just asking and suppose it's ok like this.

PR #27, opened 2020-03-26
✨(ashley) add support for unique LTI launch URL per forum

madmatah:
## Purpose

We want to have a unique LTI launch URL per forum.
We want to be able to create multiple forum per LTI context.

##  Proposal

- [x] Add an UUID to identify each forum instead of using the LTI `context_id`
- [x] Create a LTIContext model
- [x] Remove the `lti_context` and `lti_consumer` fields from the Forum model and add a manyToMany relationship to LTIContext instead
- [x] Add group management/synchronization functions inside LTIContext (and remove the `ashley.permissions.groups` module)


sandbox/dev_tools/views.py
```diff
@@ -27,7 +27,14 @@ def dev_consumer(request: HttpRequest) -> HttpResponse:
         form = LTIConsumerForm(request.POST)
         if form.is_valid():
             launch_url = request.build_absolute_uri(
-                reverse("lti_provider.views.launch")
+                reverse(
+                    "forum.lti.view",
+                    kwargs={
+                        "uuid": uuid.uuid5(
```
sampaccoud: uuid5?

tests/ashley/test_models_lticontext.py
```diff
@@ -0,0 +1,115 @@
...
+        self.assertEqual(
+            f"{base_group_name}__role__instructor", role_group_instructor.name
+        )
+
+    def test_sync_user_groups(self):
+        """Test group synchronization"""
+
+        # Create 2 LTI Contexts
+        context1 = LTIContextFactory()
+        context2 = LTIContextFactory()
```
sampaccoud:
```suggestion
        context1, context2 = LTIContextFactory.create_batch(2)
```

tests/ashley/test_models_lticontext.py
```diff
@@ -0,0 +1,115 @@
...
+        unrelated_group = Group.objects.create(name="unrelated_django_group")
+
+        # Initialize a user with no group
+        user = UserFactory()
+        self.assertEqual(0, user.groups.count())
+
+        # Sync user groups in context1 with role "student"
+        context1.sync_user_groups(user, ["student"])
+        self.assertCountEqual(
+            [context1.base_group_name, f"{context1.base_group_name}__role__student"],
```
sampaccoud: why make this to just check the count!?
Why don't you check equality? If it's because you don't know the order, maybe you can check equility of corresponding sets?

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+                login(request, user)
+                return self._do_on_login(lti_request)
+            return HttpResponseForbidden("Forbidden")
+        except LTIException:
+            return HttpResponseForbidden("Forbidden")
+
+    @abstractmethod
+    def _do_on_login(self, lti_request: LTI) -> HttpResponse:
+        """Process the request when the user is logged in via LTI"""
+
```
sampaccoud:
```suggestion
        raise NotImplementedError()
```

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+    def post(self, request, *args, **kwargs) -> HttpResponse:  # pylint: disable=W0613
+        """Handler for POST requests."""
+        lti_request = LTI(request)
+        try:
+            lti_request.verify()
+            user = authenticate(request, lti_request=lti_request)
+            if user is not None:
+                login(request, user)
+                return self._do_on_login(lti_request)
+            return HttpResponseForbidden("Forbidden")
```
sampaccoud: If you have nothing else to say, I think you can just :
```suggestion
            return HttpResponseForbidden()
```

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+class ForumLTIView(BaseLTIView):
+    """Forum view called by an LTI launch request."""
+
+    def _do_on_login(self, lti_request: LTI) -> HttpResponse:
+        """Process the request when the user is logged in via LTI"""
+
+        # Get or create the LTIContext model associated with the current LTI launch request
+        context_id = lti_request.get_param("context_id")
+        if not context_id:
+            return HttpResponseBadRequest()
```
sampaccoud: 400 errors are usually returned with an error dict?

tests/ashley/test_models_lticontext.py
```diff
@@ -0,0 +1,115 @@
...
+        # Sync user groups in context2 with multiple roles
+        context2.sync_user_groups(user, ["role1", "role2"])
+        self.assertCountEqual(
+            [
+                unrelated_group.name,
+                context1.base_group_name,
+                f"{context1.base_group_name}__role__student",
+                context2.base_group_name,
+                f"{context2.base_group_name}__role__role1",
+                f"{context2.base_group_name}__role__role2",
```
sampaccoud: Should we have a different character between the parameter name and its value?
group_name__role:role1 ? or | ? or...

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+        context, _context_created = LTIContext.objects.get_or_create(
+            remote_id=context_id, lti_consumer=lti_request.get_consumer()
+        )
+
+        # Synchronize the user groups related to the current LTI context
+        context.sync_user_groups(self.request.user, lti_request.roles)
```
sampaccoud: I think you didn't test the case of user that had more groups before the sync ie confirm that the extra group is removed?

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+
+        forum_uuid = self.kwargs["uuid"]
+        forum, forum_created = Forum.objects.get_or_create(
+            lti_id=forum_uuid,
+            type=Forum.FORUM_POST,
+            defaults={"name": lti_request.context_title, "type": Forum.FORUM_POST},
```
sampaccoud: type is already a filter parameter

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+            translation.activate(locale)
+            response.set_cookie(settings.LANGUAGE_COOKIE_NAME, locale)
+
+        return response
+
+    def _init_forum(self, forum, context):
+        """
+        Initialize a forum.
+
+        This method set default permissions according to the LTI context that
```
sampaccoud:
```suggestion
        This method sets default permissions according to the LTI context that
```

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+        initiated the creation of the forum.
+        """
+        forum.lti_contexts.add(context)
+        self._assign_perm(
```
sampaccoud: Not a big fan of shortcuts, `_assign_permission` seems a better name?

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+            forum, context.base_group(), settings.ASHLEY_DEFAULT_FORUM_BASE_PERMISSIONS
+        )
+        for role, perms in settings.ASHLEY_DEFAULT_FORUM_ROLES_PERMISSIONS.items():
+            self._assign_perm(forum, context.role_group(role), perms)
+
+    @staticmethod
+    def _assign_perm(forum, group, permissions: List[str]):
+        for perm in permissions:
```
sampaccoud: missing docstring. 
I'm surprised the linters didn't break?

src/ashley/views.py
```diff
@@ -0,0 +1,117 @@
...
+            logger.debug(
+                "Grand permission %s to group %s on forum %s",
```
sampaccoud:
```suggestion
                "Grant permission %s to group %s on forum %s",
```

src/ashley/settings.py
```diff
@@ -0,0 +1,39 @@
...
+ROLE_BASED_PERMISSIONS = {
+    "administrator": FORUM_ADMIN_PERMISSIONS,
+    "instructor": FORUM_ADMIN_PERMISSIONS,
+}
+
+settings.ASHLEY_DEFAULT_FORUM_BASE_PERMISSIONS = getattr(
+    settings, "ASHLEY_DEFAULT_FORUM_BASE_PERMISSIONS", FORUM_BASE_PERMISSIONS
+)
+
+settings.ASHLEY_DEFAULT_FORUM_ROLES_PERMISSIONS = getattr(
```
sampaccoud:
```suggestion
DEFAULT_FORUM_ROLES_PERMISSIONS = getattr(
```

src/ashley/settings.py
```diff
@@ -0,0 +1,39 @@
...
+    "can_approve_posts",
+    "can_reply_to_locked_topics",
+]
+
+ROLE_BASED_PERMISSIONS = {
+    "administrator": FORUM_ADMIN_PERMISSIONS,
+    "instructor": FORUM_ADMIN_PERMISSIONS,
+}
+
+settings.ASHLEY_DEFAULT_FORUM_BASE_PERMISSIONS = getattr(
```
sampaccoud: you shouldn't modify the settings object. Why not define your own variable and import it where you need it?
```suggestion
DEFAULT_FORUM_BASE_PERMISSIONS = getattr(
```

src/ashley/models.py
```diff
@@ -64,3 +70,100 @@ class Meta:
...
 User = model_factory(AbstractUser)
+
+
+class AbstractLTIContext(Model):
+    """Abstract LTI Context model for Ashley."""
+
+    GROUP_PREFIX = "cg"
+    GROUP_DELIMITER = "__"
+
+    remote_id = models.CharField(max_length=150, null=False, blank=False, db_index=True)
```
sampaccoud: why not `lti_id`?

src/ashley/models.py
```diff
@@ -64,3 +70,100 @@ class Meta:
...
+
+    lti_consumer = models.ForeignKey(
+        LTIConsumer, on_delete=models.CASCADE, null=False, db_index=True
```
sampaccoud: I find it dangerous and probably useless to cascade deletion:
```suggestion
        LTIConsumer, on_delete=models.RESTRICT, null=False, db_index=True
```

src/ashley/models.py
```diff
@@ -64,3 +70,100 @@ class Meta:
...
+    )
+
+    def base_group(self) -> Group:
```
sampaccoud: For such method getters, I think you should give names that defer from property names. 
I propose naming them : `get_*` so here:
```suggestion
    def get_base_group(self) -> Group:
```

src/ashley/models.py
```diff
@@ -64,3 +70,100 @@ class Meta:
...
+        return self._get_or_create_group(self.base_group_name)
+
+    @property
+    def base_group_name(self) -> str:
+        """
+        Get the name of the base group.
+        """
+        return self.GROUP_DELIMITER.join([self.GROUP_PREFIX, str(self.id)])
+
+    def role_groups(self, roles: List[str]) -> List[Group]:
```
sampaccoud:
```suggestion
    def get_role_groups(self, roles: List[str]) -> List[Group]:
```

src/ashley/models.py
```diff
@@ -64,3 +70,100 @@ class Meta:
...
+        """
+        Get the Django Groups corresponding to a list of LTI Context Roles
+
+        All LTI users authenticated within this LTI context having these roles
+        should be in these groups.
+        """
+        return list(map(self.role_group, roles,))
+
+    def role_group(self, role: str) -> Group:
```
sampaccoud:
```suggestion
    def get_role_group(self, role: str) -> Group:
```

src/ashley/machina_extensions/forum/migrations/0014_convert_lti_context.py
```diff
@@ -0,0 +1,28 @@
+from django.db import migrations
```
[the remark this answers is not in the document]
madmatah: Why ? I followed the official guidelines for my use case :
https://docs.djangoproject.com/en/3.0/howto/writing-migrations/#migrations-that-add-unique-fields

I'll try to group them in a single file.

sandbox/dev_tools/views.py
madmatah: I use uuid5 + NAMESPACE_DNS to generate a consistent uuid based on a string.
This is to simplify the dev tools interface..

tests/ashley/test_models_lticontext.py
madmatah: Thanks for the tip :+1:

tests/ashley/test_models_lticontext.py
madmatah: Despite its misleading name, this is exactly the purpose of [assertCountEqual](https://docs.python.org/3.2/library/unittest.html#unittest.TestCase.assertCountEqual) :

> **assertCountEqual(first, second, msg=None)**
> Test that sequence first contains the same elements as second, regardless of their order. When they don’t, an error message listing the differences between the sequences will be generated.

src/ashley/machina_extensions/forum/migrations/0014_convert_lti_context.py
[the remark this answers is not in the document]
sampaccoud: if you delete all you migrations and replay makemigrations you should get them all under one file...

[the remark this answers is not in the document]
sampaccoud: Actually, maybe I'm wrong that it's better to make everything in one file :thinking: each migration file is run atomically. So my point is that you either want all the migrations in your PR to pass or all to fail.

tests/ashley/test_models_lticontext.py
sampaccoud: oh wow. I learnt something. shitty name indeed.

src/ashley/machina_extensions/forum/migrations/0014_convert_lti_context.py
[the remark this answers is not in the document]
madmatah: Not in this case, I had to write some of them manually.

1) To add an uuid with unique=True, I followed the instructions described in the link I posted above (migration 16 to 18).
2) To convert two fields into a many-to-many (migration 13 to 15)

I'll try to write them in a single operations array in the same migration file.

tests/ashley/test_models_lticontext.py
madmatah: Yep... I was named `assertSameElements` before and has been deprecated in 3.2 in flavor of this new name. I don't understand why..

src/ashley/views.py
madmatah: I'll add it, but to understand better, why is it required even if I use the `@abstractmethod` decorator ?

madmatah: And another question on this class : do you think I should move `BaseLTIView` in the `lti_provider` module ? I think it might be a better place to put this base class.

src/ashley/views.py
madmatah: This case is tested here : 
https://github.com/openfun/ashley/blob/feature/unique-lti-link/tests/ashley/test_models_lticontext.py#L93

The user in the context2 has previously roles `role1` and `role2`.
After the sync, it is not a member of these groups anymore.

tests/ashley/test_models_lticontext.py
madmatah: Ok, go for `:`

src/ashley/views.py
madmatah: Fixed :+1:
The linter does not require docstring on protected methods.

src/ashley/models.py
madmatah: I agree with you, I changed it to `on_delete=models.PROTECT`

src/ashley/views.py
sampaccoud: 1. I'm sure at all but I don't feel good leaving this method with an empty body and returning None when it is in fact not usable as such. I'm not sure why you needed to derive from `ABC` in the first place?
2. I agree the `BaseLTIView` as its place in the `lti_provider` module!

madmatah: 1. I thought ABC was the way to make abstract classes
2. Ok thanks, i'll move it, then !

src/ashley/settings.py
madmatah: These are settings that ashley operators can override.
I use the same [pattern](https://stackoverflow.com/a/8429282) than in `lti_provider` and this provides default values for these settings.

The constants defined here are only for readability purpose.

src/ashley/machina_extensions/forum/migrations/0014_convert_lti_context.py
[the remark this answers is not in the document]
madmatah: Ok it is now in in a single file, I tested it and it works  :heavy_check_mark:

src/ashley/views.py
madmatah: Done :heavy_check_mark:

src/ashley/views.py
madmatah: I don't understand your comment ?
Do you have an example ?

sampaccoud: I found this example: https://github.com/openfun/marsha/blob/master/src/backend/marsha/core/api.py#L52

sampaccoud: Looking at test is easier to see what the errors dict looks like:
https://github.com/openfun/marsha/blob/master/src/backend/marsha/core/tests/test_serializers_update_state.py#L55

PR #35, opened 2020-04-20
✨(ashley) render emojis with emojione

madmatah:
## Purpose

Emojis are rendered with emojione in ashley WISYWIG editor. But when a
forum post is displayed, emojis are rendered with the native font of
the user's browser. In consequence, the rendering is inconsistent and
sometime emojis are not displayed at all.

## Proposal

Use emojione to render emojis everywhere.

### DIFF
--- a/src/frontend/js/ashley.ts
+++ b/src/frontend/js/ashley.ts
@@ -1,7 +1,12 @@
 import * as AshleyEditor from './components/AshleyEditor';
+import { renderEmojis } from './utils/emojis';
 
 // expose some modules to the global window object
 declare var window: any;
 window.Ashley = {
   Editor: AshleyEditor,
 };
+
+document.addEventListener('DOMContentLoaded', event => {
+  renderEmojis();
+});
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -46,12 +46,14 @@
     "webpack-cli": "3.3.11"
   },
   "dependencies": {
+    "@types/emojione": "^2.2.6",
     "draft-js": "0.11.5",
     "draft-js-anchor-plugin": "2.0.3",
     "draft-js-buttons": "2.0.2",
     "draft-js-emoji-plugin": "2.1.3",
     "draft-js-plugins-editor": "3.0.0",
     "draft-js-static-toolbar-plugin": "3.0.1",
+    "emojione": "4.5.0",
     "es6-shim": "0.35.5",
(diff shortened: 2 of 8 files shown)


src/frontend/js/ashley.ts
```diff
@@ -1,7 +1,12 @@
 import * as AshleyEditor from './components/AshleyEditor';
+import { renderEmojis } from './utils/emojis';
 
 // expose some modules to the global window object
 declare var window: any;
 window.Ashley = {
   Editor: AshleyEditor,
 };
+
+document.addEventListener('DOMContentLoaded', async event => {
```
lunika, line 10: Why the callable is `async` ? There is no async code used here

src/frontend/webpack.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   // Currently, @babel/preset-env is unaware that using import() with Webpack relies on Promise internally.
   // Environments which do not have builtin support for Promise, like Internet Explorer, will require both
   // the promise and iterator polyfills be added manually.
-  entry: ['./js/ashley.ts'],
+  entry: ['regenerator-runtime/runtime', './js/ashley.ts'],
```
lunika, line 10: The comment before the `entry` setting is not accurate anymore.

src/frontend/package.json
```diff
@@ -46,14 +46,17 @@
...
     "draft-js-buttons": "2.0.2",
     "draft-js-emoji-plugin": "2.1.3",
     "draft-js-plugins-editor": "3.0.0",
     "draft-js-static-toolbar-plugin": "3.0.1",
+    "emojione": "4.5.0",
     "es6-shim": "0.35.5",
     "react": "16.13.1",
-    "react-dom": "16.13.1"
+    "react-dom": "16.13.1",
+    "regenerator-runtime": "0.13.5"
```
lunika, line 60: Why adding this package ?

src/frontend/js/ashley.ts
madmatah: Oops, I thought it was the new way to do that :smile: 
I'll remove the `async` part.

src/frontend/package.json
madmatah: It was to avoid the error «ReferenceError: regeneratorRuntime is not defined» caused by the async callback. I'll also remove it.

PR #36, opened 2020-04-20
✨(ashley) add support for Moodle in ashley's LTI authentication backend

madmatah:
## Purpose

I want to use Ashley as an external tool on Moodle, but I get a 403 with the following error : 

> Unable to find param lis_person_sourcedid in LTI request

## Proposal 

The remote username (on the LTI consumer side) is sent to LTI Providers via various LTI parameters, depending on the LTI Consumer that initiates the launch request.
In OpenEdx, the parameter is `lis_person_sourcedid`.
In Moodle, the parameter is `ext_user_username`.

This PR adds support for `ext_user_username`.

As a result, we can use Ashley in Moodle :

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ## Added
 
  - render emoji with emojione on forum posts
+ - add support for Moodle in ashley's LTI authentication backend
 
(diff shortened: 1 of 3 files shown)


sampaccoud:
@madmatah I think it deserves a test?

madmatah:
> @madmatah I think it deserves a test?

@sampaccoud That's true ! thanks for the reminder :wink:  I added a test.

PR #39, opened 2020-04-21
♻️(front) transform ashleyeditor component in functional component

lunika:
## Purpose

We want to rewrite the ashleyeditor component in functional way using
hooks. To keep toolbar in top of the editor we use a css hack forcing
the top position on top on the editor.

## Proposal

- [x] refactor `<AshleyEditor />` in functional component using hooks

### DIFF
--- a/src/frontend/js/components/AshleyEditor/index.tsx
+++ b/src/frontend/js/components/AshleyEditor/index.tsx
@@ -10,14 +10,11 @@ import {
   UnderlineButton,
   UnorderedListButton,
 } from 'draft-js-buttons';
-import createEmojiPlugin, {
-  EmojiPlugin,
-  EmojiPluginConfig,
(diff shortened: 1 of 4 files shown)


src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -27,81 +24,92 @@ interface MyEditorProps {
...
+    }
+    return EditorState.createEmpty();
+  });
+
+  const toolbarRef = useRef(null as HTMLDivElement | null);
+  const editorContainerRef = useRef(null as HTMLDivElement | null);
+  const editorRef = useRef(null as PluginEditor | null);
+
+  // Instantiate plugins in a state to avoid instantiation on every render
+  // Emoji plugin
```
madmatah, line 43:
```suggestion
```

PR #50, opened 2020-04-27
✨(front) add button to insert a link in AshleyEditor component

madmatah:
## Purpose

As a Ashley user, I want to insert links in my forum posts.

## Proposal

Add a button in `AshleyEditor`'s toolbar, using [draft-js-anchor-plugin](https://www.draft-js-plugins.com/plugin/anchor).

### DIFF
--- a/src/frontend/js/components/AshleyEditor/index.tsx
+++ b/src/frontend/js/components/AshleyEditor/index.tsx
@@ -22,6 +22,7 @@ interface MyEditorProps {
   placeholder?: string;
   target: HTMLInputElement;
   emojiConfig?: EmojiPluginConfig;
+  linkPlaceholder?: string;
 }
 
 const AshleyEditor = (props: MyEditorProps) => {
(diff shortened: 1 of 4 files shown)


src/frontend/scss/objects/_editor.scss
```diff
@@ -0,0 +1,55 @@
+.ashley-editor-widget {
+  border: 1px solid #ced4da;
```
jmaupetit, line 2: Where does those colors come from? Shouldn't we use variables instead?

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -22,6 +22,7 @@ interface MyEditorProps {
   placeholder?: string;
   target: HTMLInputElement;
   emojiConfig?: EmojiPluginConfig;
+  linkPlaceholder?: string;
```
lunika, line 25: Is this props really optional ? The props are editable and can be changed/removed ?

src/frontend/scss/objects/_editor.scss
madmatah: Sorry for the delay, is it better now ? :slightly_smiling_face:

src/frontend/js/components/AshleyEditor/index.tsx
madmatah: Sorry for the delay.
I made this props optional because the draftjs plugin provides a default placeholder.
One the editor is initialized, this cannot be changed/removed.

PR #56, opened 2020-05-13
💄(front) coalesce SASS color variables between machina and bootstrap

madmatah:
## Purpose

The SASS color variables defined in django-machina theme is not homogeneous with colors defined in bootstrap.

## Proposal

- [x] replace color variables defined in django-machina theme with similar variables from bootstrap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
  - build the forum CSS theme inside ashley instead of using the compiled CSS
    file distributed with django-machina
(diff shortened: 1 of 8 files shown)


jmaupetit:
So basically the idea is to stick to bootstrap variable names rather than machina?

madmatah:
> So basically the idea is to stick to bootstrap variable names rather than machina?

Yes the idea is to remove colors defined by machina, in order to have a simple color palette to start with. And also, avoid conflicts between variables defined by bootstrap and machina with a different value (e.g. `$blue`).

PR #88, opened 2020-09-17
💥(lti) replace lti_provider app with the `django-lti-toolbox` dependency

### LINKED ISSUE #64: Use django-lti-toolbox library to replace old code originally integrated in Ashley
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
The [django lti toolbox](https://github.com/openfun/django-lti-toolbox) project was created to share the code related to handling LTI requests, with other similar projects that are proposing an app as LTI provider.

The code of this library was extracted from Ashley, so we can now make use of this new library and remove the corresponding code from Ashley.

**Describe the solution you'd like**
- [ ] Add a dependency to [django lti toolbox](https://github.com/openfun/django-lti-toolbox) in Ashley Python requirements
- [ ] Remove old code that is now useless
- [ ] Adapt tests to pass and secure the way we use LTI

**Discovery, Documentation, Adoption, Migration Strategy**
While preparing your pull request, follow guidelines in our [handbook](https://openfun.gitbooks.io/handbook/content/) and look at examples of pull requests that have already been merged to see how we do it.

madmatah:
## Purpose

We want to use `django-lti-toolbox` dependency instead of the
`lti_provider` app included in this project. This is a copy of this
app, it has been externalized in order to use it in multpile projects.

## Proposal

- [x] Add a dependency to django lti toolbox in Ashley Python requirements
- [x] Remove old code that is now useless
- [x] Adapt tests to pass and secure the way we use LTI
- [x] Adapt database migrations && propose upgrade instructions

## Migrations / Upgrade instructions

:warning:  The migration part of this refactoring is tricky because of the following reasons : 

- former `lti_provider` app has been renamed to `lti_toolbox`
- tables created by `lti_provider` app and `lti_toolbox` have the same name
- The first migration file introduced by ashley depends on [`lti_provider`, `0001_initial`].

I didn't find a way to handle this with standard migration scripts without breaking `ashley` or `lti_toolbox`.

That's why I propose to execute the following instructions to upgrade :

1) Replace `lti_provider` with `lti_toolbox` in your INSTALLED_APPS.
2) Execute the following SQL query in your database :
   `UPDATE django_migrations SET app='lti_toolbox' WHERE
    app='lti_provider' AND name='0001_initial';`

Since `lti_toolbox` and `lti_provider` migrations are strictly the same, it works in all cases.

If you have a better solution to offer, I'm interrested ! :-) 

Fixes #64


sampaccoud:
@madmatah Have you tried running the schemamigrations after changing the app name? It may catch what you are doing and propose the change. If not, it will be better to write your own migration code to do what you propose to do in SQL so that the migration can be done by "just" running the `migrate` command.

madmatah:
> @madmatah Have you tried running the schemamigrations after changing the app name? It may catch what you are doing and propose the change. If not, it will be better to write your own migration code to do what you propose to do in SQL so that the migration can be done by "just" running the `migrate` command.

What are you refering to when you write about «running the schemamigrations» ? I cannot find a management command with this name.

About writing the SQL in a migration command, the thing is : 

The `migrate` command can't be executed because of the following error : 
`Migration ashley.0002_auto_20200325_1651 dependencies reference nonexistent parent node ('lti_provider', '0001_initial')`
This is due to the fact that the former `lti_provider` app has been removed (files + entry in INSTALLED_APPS).

That's why I patched the dependencies name in the migration files, to replace `lti_provider` with `lti_toolbox`.

 It allows "fresh installations" of ashley to execute the migrations without error.
For previous installations of ashley, the migration state is inconsistent after this patch.

If you execute the migrate commande, you have the following error : 

`Migration ashley.0001_initial is applied before its dependency lti_toolbox.0001_initial on database 'default'.`

That's why my proposition is to patch the migration state.
And that cannot be done with a migration command.

madmatah:
To simplify Ashley operators' experience, I added a `UPGRADE.md` file and a migration script in the `upgrades` directory.

UPGRADE.md
```diff
@@ -0,0 +1,23 @@
+# Upgrade instructions
+
+For most upgrades, you just need to run the django migrations with
+the following command inside your docker container :
+
+`python manage.py migrate`
```
jmaupetit, line 6: Make it a `code` block:

```suggestion
```python manage.py migrate```
```

UPGRADE.md
```diff
@@ -0,0 +1,23 @@
...
+
+
+(Note : in your development environment, you can run `bin/manage migrate`)
```
jmaupetit, line 9: Or `make migrate`?

UPGRADE.md
```diff
@@ -0,0 +1,23 @@
+# Upgrade instructions
+
+For most upgrades, you just need to run the django migrations with
+the following command inside your docker container :
```
jmaupetit, line 4:
```suggestion
the following command inside your docker container:
```

UPGRADE.md
```diff
@@ -0,0 +1,23 @@
...
+
+`python manage.py migrate`
+
+
+(Note : in your development environment, you can run `bin/manage migrate`)
+
+
+Some upgrades need special upgrade instructions.
+They will be documented in this file :
```
jmaupetit, line 13:
```suggestion
They will be documented in this file:
```

UPGRADE.md
```diff
@@ -0,0 +1,23 @@
...
+
+
+### Ashley 1.0.0-beta.3
+
+To upgrade from any previous version of ashley to version 1.0.0-beta.3, you must execute
+the following instructions :
```
jmaupetit, line 19:
```suggestion
the following instructions:
```

jmaupetit (approved):

PR #107, opened 2020-11-26
💄(backend) update search form template

madmatah:
## Purpose

We don't want to display the default machina navbar in the header. 

## Proposal

- [x] Remove the navbar
- [x] Add a search form below the breadcrumbs

Screenshot from 2020-11-26 20-54-18


src/ashley/templates/board_base.html
```diff
@@ -7,25 +7,37 @@
...
+        <div class="row">
+            <div class="col-12">
+                {% block breadcrumb %}{% include "partials/breadcrumb.html" %}{% endblock breadcrumb %}
+            </div>
+        </div>
+
+        {% block search %}
+        <div class="row">
+            <form class="form-inline my-2 ml-3 my-lg-0" role="search" action="{% url 'forum_search:search' %}">
+                <input class="form-control form-control-sm mr-sm-2" type="search" placeholder="Search" aria-label="Search" name="q">
```
jbpenrath, line 22: What do you think about adding a `required` attribute to the search field to prevent to submit the form with an empty query param?

src/ashley/templates/board_base.html
```diff
@@ -7,25 +7,37 @@
...
+                <button class="btn btn-primary btn-sm my-2 my-sm-0" type="submit">{% trans "Search" %}</button>
+                <a class="small pl-2" href="{% url 'forum_search:search' %}">{% trans "Advanced search" %}</a>
+            </form>
+        </div>
+        {% endblock search %}
+
+        <div class="row">
+            <div class="col-12">
+                <br/>
```
jbpenrath, line 31: I would use a padding-top (`pt-3` ?) instead of `<br />`

jbpenrath (approved):
✅ Just nitpicks

src/ashley/templates/board_base.html
madmatah: I think it's a very good suggestion! :pray:

madmatah:
@jbpenrath Thank you very much for your review, i addressed your feedbacks.

Issue #110, opened 2020-12-01
DraftJS editor resizing / toolbar rendering issue


madmatah:
## Bug Report

**Expected behavior/code**

When an user enters a message that does not fit vertically in the WYSIWYG editor, the editor should expand its height.

**Actual Behavior**

The toolbar is incorrectly rendered when the resizing happens. (see screenshot below)

**Steps to Reproduce**
1. On any forum, click on «New topic»
2. Enter a message that is at least 6 lines long.
3. And then the bug happens!

**Additional context/Screenshots**

![draftjs-editor](https://user-images.githubusercontent.com/2926/100752958-7ba8f900-33e9-11eb-998c-4d3e5974104d.gif)

PR #111, opened 2020-12-02
🐛(front) fix Draft.js editor resize issue with static toolbar

madmatah:
## Purpose

We used a CSS hack to be able to place the Draft.js static toolbar
plugin above the editor, because it was not possible to place directly
the static-toolbar component above the editor.

This hack does not handle well the resizing of the editor, because of
the way it computes the height of the editor. It is one time behind
the reality.

Here is the result, as explained in #110 : 

## Proposal

Use absolute positionning for the toolbar instead of relative.
It avoids computing the height of the editor to position it.

### DIFF
--- a/src/frontend/js/components/AshleyEditor/index.tsx
+++ b/src/frontend/js/components/AshleyEditor/index.tsx
@@ -36,8 +36,6 @@ const AshleyEditor = (props: MyEditorProps) => {
     return EditorState.createEmpty();
   });
 
-  const toolbarRef = useRef(null as HTMLDivElement | null);
-  const editorContainerRef = useRef(null as HTMLDivElement | null);
   const editorRef = useRef(null as PluginEditor | null);
 
   // Instantiate plugins in a state to avoid instantiation on every render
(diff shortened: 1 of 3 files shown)


madmatah:
Thanks to @jbpenrath suggestions, here is a much better and simpler proposal.

src/frontend/scss/objects/_editor.scss
```diff
@@ -1,3 +1,7 @@
+.ashley-editor-wrapper {
+  position: relative;
```
jbpenrath, line 2:
```suggestion
  position: relative;
  display: flex;
  flex-direction: column-reverse;
```

src/frontend/scss/objects/_editor.scss
```diff
@@ -8,7 +12,9 @@
 }
 
 .ashley-editor-toolbar {
-  position: relative;
+  position: absolute;
```
jbpenrath, line 15:
```suggestion
  position: sticky;
  top: 0;
  width: 100%;
  z-index: 2;
```

jbpenrath (commented):
Like a flash, I like a better idea!

madmatah:
@jbpenrath Thank you very much for this suggestion which solves all our concerns :+1:

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -86,19 +85,17 @@ const AshleyEditor = (props: MyEditorProps) => {
...
-          toolbarRef.current
-            ? {
-                top: `${toolbarRef.current.offsetHeight}px`,
-              }
-            : {}
-        }
-      >
+    <div
+      className="ashley-editor-wrapper"
+      style={
```
jbpenrath, line 90: By using `flex-direction: column-reverse`, you do not need anymore to add a padding to the wrapper.

jbpenrath (approved):

src/frontend/js/components/AshleyEditor/index.tsx
madmatah: That's right. :pray:

PR #118, opened 2020-12-10
✨(sandbox) add and configure sentry SDK

madmatah:
## Purpose

We want to monitor Ashley's sandbox with sentry, like in our other projects.

## Proposal

- [x] Add sentry_sdk dependency for the sandbox
- [x] Update sandbox settings


jmaupetit (approved):

sandbox/settings.py
```diff
@@ -232,6 +255,44 @@ class Base(Configuration):
...
+        This is the place where you can configure settings that require other
+        settings to be loaded.
+        """
+        super().post_setup()
+
+        # The SENTRY_DSN setting should be available to activate sentry for an environment
+        if cls.SENTRY_DSN is not None:
+            sentry_sdk.init(
+                dsn=cls.SENTRY_DSN,
+                environment=cls.ENVIRONMENT,
```
lunika, line 289: I'm pretty sure this is not working. You should test I think you will have something like `<property object at 0x7f47aebb9c50>`.
you should use 

```suggestion
                environment=return cls.__name__.lower(),
```

sandbox/settings.py
```diff
@@ -232,6 +255,44 @@ class Base(Configuration):
...
+                release=cls.RELEASE,
```
lunika, line 290: Same here, you should use

```suggestion
               release=get_release(),
```

sandbox/settings.py
```diff
@@ -232,6 +255,44 @@ class Base(Configuration):
     MACHINA_PROFILE_AVATARS_ENABLED = False
     MACHINA_USER_DISPLAY_NAME_METHOD = "get_public_username"
 
+    # Sentry
+    SENTRY_DSN = values.Value(None, environ_name="SENTRY_DSN")
+
+    # pylint: disable=invalid-name
+    @property
+    def ENVIRONMENT(self):
```
lunika, line 263: As you are not reusing this property, you can directly  you its value in the `post_setup` function

sandbox/settings.py
```diff
@@ -232,6 +255,44 @@ class Base(Configuration):
...
+        """Environment in which the application is launched."""
+        return self.__class__.__name__.lower()
+
+    # pylint: disable=invalid-name
+    @property
+    def RELEASE(self):
```
lunika, line 269: Ditto

sandbox/settings.py
madmatah: You're right! Thanks.
I fixed it.

lunika (approved):

PR #130, opened 2021-01-06
📝 (readme) add access admin and troubleshooting elasticsearch ser…

carofun:
## Purpose

Adding instructions to the readme file to ease the use of the project 

## Proposal

- Add the url of the backoffice and how to create a superuser.
- Add the troubleshooting that mac users have with a default docker configuration with ElasticSearch service

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,3 +62,29 @@ $ make run
 
 You should be able to test the forum at [http://localhost:8090/dev/](http://localhost:8090/dev/).
 This is the development view that emulates a LTI consumer.
+
+## Django admin
+
+You can access the Django admin site of the forum at [http://localhost:8090/admin/](http://localhost:8090/admin/)
+
+You first need to create a superuser account. 
+
+```bash
+$ make superuser
+```
+
+## Troubleshooting
+
+### ElasticSearch service out with an Exit 137
+
+```bash
+$ docker-compose ps
+```
+> ashley_elasticsearch_1   /docker-entrypoint.sh elas ...   Exit 137
+
+This is due to a lack of memory.
+N.B. for Mac users, the default Docker's configuration is also 2GB which means that Elasticsearch will automatically exit.
+
+To fix this issue, you need to upgrade your default configuration. 
+
+


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
  - upgrade django-machina from 1.1.3 to 1.1.4
+ - update readme instructions on accessing backoffice
```
lunika, line 14: IMO there is no need to update the changelog for this kind of modification.

README.md
```diff
@@ -61,4 +61,28 @@ $ make run
...
+
+You first need to create a superuser account. 
+
+```bash
+$ bin/superuser
+```
+
+## Troubleshooting
+
+ElasticSearch service is always down with an Exit 137 State
```
lunika, line 78: Is it also related to the issue we have on richie ? If it's the case you can maybe add more information on how to deal with this error for non mac users: https://richie.education/docs/docker-development#elasticsearch-service-is-always-down

README.md
```diff
@@ -61,4 +61,28 @@ $ make run
...
+This is the development view that emulates a LTI consumer.
+
+## Backoffice
+
+You can access the backoffice of the forum at [http://localhost:8090/admin/](http://localhost:8090/admin/)
+
+You first need to create a superuser account. 
+
+```bash
+$ bin/superuser
```
lunika, line 73: Interesting to see there is no make rule for this task. In all django projects we have a `superuser` task is responsible of this. But maybe we already discuss of this point with @madmatah ?

README.md
```diff
@@ -61,4 +61,28 @@ $ make run
 ```
 
 You should be able to test the forum at [http://localhost:8090/dev/](http://localhost:8090/dev/).
-This is the development view that emulates a LTI consumer.
\ No newline at end of file
+This is the development view that emulates a LTI consumer.
+
+## Backoffice
```
madmatah, line 66: I would replace the term «backoffice» with «Django admin site» or «Django admin» since it seems to be its official name.

README.md
carofun: I thought about it but actually I didn't have the error on ritchie because it's not the same elasticsearch version so I'm not quiet sure it's exactly relevant

madmatah, line 78: We might have multiple solutions to provide in this section, so I would add this in a sub-section  : 

```suggestion
### ElasticSearch service is always down with an Exit 137 State
```

README.md
carofun: We've actually noticed it and as it's my first PR let it this way, to see what could come out of it :-)

CHANGELOG.md
carofun: How do we deal with circle ci rule then ?

jbpenrath: The job `check-changelog` success is not required to merge the PR.

README.md
carofun: thank you for your feedback I will do it

CHANGELOG.md
carofun: Exact and regarding the last commit it is not needed, I'll remove it.

README.md
madmatah: There was a time when I had a preference for the `bin/*` utilities rather than the make targets ;)
But I agree with you @lunika , to be consistent we should add a `superuser` target in the Makefile and remove this script, in another PR.

README.md
lunika: Yes but it's a little bit long for a title. Maybe just keeping `ElasticSearch service is always down`

carofun: To get around it I've change the wording, I've added a N.B and as I don't give concrete solutions for mac users, linux ones won't miss one either. In the Richie doc, we can see error in logs so I didn't want to copy this part because I didn't go through it. The command line for linux is probably right to use but then I will need to be fare to explain how to change it for mac users as well.

carofun: Regarding the error in Richie I thought it would be good to specify Exit 137 as it surprised us not to find anything in logs

carofun: It is now ### ElasticSearch service out with an Exit 137
hope it's good to everybody

README.md
carofun: The task has been added !

madmatah (approved):

lunika (approved):

PR #131, opened 2021-01-07
✨(makefile) create a superuser task

carofun:
## Purpose

To propose an easy way to create an user for the Django admin

## Proposal

Add an entry point in the makefile so that we can use the command 
> make superuser


sampaccoud (approved):

madmatah (approved):
Just one thing : can you remove the `bin/superuser` script since it is no longer necessary ?
Otherwise :

PR #132, opened 2021-01-07
🐛(search) fix serialization error in search index update

madmatah:
## Purpose

The haystack command `update_index` crashes while indexing posts with
a serialization error. It happens when a message is posted by a user
having an empty public_username field.

This is due to the fact that in this specific case, we are returning a
default public_username ("Anonymous") translated with gettext_lazy.
And lazy proxy values are not directly serializable into JSON.

## Proposal

- [x] do not index the default «Anonymous» username
- [x] update the search result template to display the default username if not specified in the search result

### DIFF
--- a/src/ashley/models.py
+++ b/src/ashley/models.py
@@ -48,6 +48,10 @@ class AbstractUser(DjangoAbstractUser):
 
     def get_public_username(self):
         """Getter for the public username of the user."""
+        return self.public_username
+
+    def get_public_username_with_default(self):
+        """Getter for the public username of the user."""
         return self.public_username or _("Anonymous")
 
     class Meta:
--- a/src/ashley/templates/forum_search/search.html
+++ b/src/ashley/templates/forum_search/search.html
@@ -1,0 +1,120 @@
+{% extends 'board_base.html' %}
+{% load i18n %}
+{% load highlight %}
+
+{% block sub_title %}{% trans "Search" %}{% endblock sub_title %}
+
+{% block content %}
+{% if not query %}
+<div class="row">
+  <div class="col-12">
+    <h1>{% trans "Search" %}</h1>
+  </div>
+</div>
+<div class="row">
+  <div class="col-12">
+    <div class="card post-edit">
+      <div class="card-header">
+        <h3 class="m-0 card-title h5">{% trans "Search forums" %}</h3>
+      </div>
+      <div class="card-body">
+        <form method="get" action="." class="form">
+          {% include "partials/form_field.html" with field=form.q %}
+          <div class="checkbox">
+            <label for="{{ form.search_topics.auto_id }}">
+              {{ form.search_topics }}
+              {{ form.search_topics.label }}
+            </label>
+          </div>
+          {% include "partials/form_field.html" with field=form.search_poster_name %}
+          {% include "partials/form_field.html" with field=form.search_forums %}
+          <div class="form-actions">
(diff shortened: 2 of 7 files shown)


tests/ashley/test_forum_search.py
```diff
@@ -399,5 +400,33 @@ def test_forum_search_restrict_poster_topic_forums(self):
         self.assertContains(response, post1.topic.slug)
         self.assertContains(response, post5.topic.slug)
 
+    def test_forum_search_empty_public_username(self):
```
sampaccoud, line 403: I'm wondering whether it's a good thing to index "anonymous" as the user's name.... Shouldn't we index a null value in this case?

madmatah: I was wondering the same thing :)

The thing is that we have to use the `get_forum_member_display_name()` shortcut in the `PostIndex` definition to get the username, because its behavior is customizable with the  `USER_DISPLAY_NAME_METHOD` machina setting.

The «Anonymous» default value is specific to our own implementation of this method and I think we should not make an exception to handle our use case here. I see 3 solutions here :

- Keep it like this and index the «Anonymous» username. 
- Add a new setting and shortcut function to get the username to index (with a default value that calls get_forum_member_display_name() )
- update our `get_public_username()` method on our User model to return a null value. But it implies to update the template helper function to handle the default «Anonymous» value.

WDYT ?

sampaccoud: I would say 2 or 3 but I can't tell which one is better from where I stand :stuck_out_tongue:

madmatah: Ok :) I'll do that in another PR then.

madmatah:
@sampaccoud I made the changes to not index the default «Anonymous» username in elasticsearch.
Also added tests.
And I had to override the search result template to display the default «Anonymous» username when the `poster_name` is not defined in the search result.

Your review on this part would we appreciated :slightly_smiling_face: :pray:

src/ashley/templates/forum_search/search.html
```diff
@@ -0,0 +1,120 @@
+{% extends 'board_base.html' %}
```
sampaccoud, line 1: note that we could propose django-machina to add {% block %} sections in their templates so we can override them with more granularity.

src/ashley/models.py
```diff
@@ -47,6 +47,10 @@ class AbstractUser(DjangoAbstractUser):
     REQUIRED_FIELDS: List[str] = []
 
     def get_public_username(self):
+        """Getter for the public username of the user."""
+        return self.public_username
+
+    def get_public_username_with_default(self):
```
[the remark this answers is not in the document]
madmatah: This method is used in the [sandbox settings](https://github.com/openfun/ashley/pull/132/files#diff-6149e0bb82135c43ba78910a0c8e83c4c6a6204a9900940574abe6b2ca53a23fR256).

The `default` template filter is only used in the search results template, because it does not receive a User model but a document from elasticsearch.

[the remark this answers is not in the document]
sampaccoud: oh. May I argue that it is not dry?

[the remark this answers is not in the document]
madmatah: Yes you may :smile: 
It's about the "Anonymous" string, or more than that ?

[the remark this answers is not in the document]
sampaccoud: Ok I see, not easy to do otherwise. Fair enough.

sampaccoud (approved):

lunika (approved):

Issue #67, opened 2020-06-24
Add quotes to the draftjs editor


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
The input field to add a message is based on the [draftjs editor](https://draftjs.org/) and [draftjs plugins](https://www.draft-js-plugins.com/). We want to be able to add a quote via this editor.

<img src="https://user-images.githubusercontent.com/1427165/85526158-0a7b4380-b60a-11ea-83a8-ad96a64b63ef.png" width="500" />

**Describe the solution you'd like**
Add a "quote" button to the editor's toolbar that allows formatting part of a message as a quote:

<img src="https://user-images.githubusercontent.com/1427165/85527167-ebc97c80-b60a-11ea-8452-1527d696afcc.png" width="30" />

**Discovery, Documentation, Adoption, Migration Strategy**
While preparing your pull request, follow guidelines in our [handbook](https://openfun.gitbooks.io/handbook/content/) and look at examples of pull requests that have already been merged to see how we do it.

PR #133, opened 2021-01-07
✨(front) add button to insert quotes to the draftjs editor

carofun:
## Purpose
The input field to add a message is based on the draftjs editor and draftjs plugins. We want to be able to add a quote via this editor.

Based on issue https://github.com/openfun/ashley/issues/67

## Proposal

Add a "quote" button to the editor's toolbar that allows formatting part of a message as a quote
Capture d’écran 2021-01-07 à 16 48 18
teest

### DIFF
--- a/src/frontend/scss/objects/_conversation_topic_detail.scss
+++ b/src/frontend/scss/objects/_conversation_topic_detail.scss
@@ -38,6 +38,10 @@
         }
       }
 
+      blockquote {
+        @include blockquote;
+      }
+
       .post-signature {
         margin-top: 20px;
         margin-bottom: 20px;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
  - upgrade django-machina from 1.1.3 to 1.1.4
+ - add button to insert quotes in the draftjs editor
 
 ## [1.0.0-beta.4] - 2020-12-22
 
--- a/src/frontend/js/components/AshleyEditor/index.tsx
+++ b/src/frontend/js/components/AshleyEditor/index.tsx
@@ -9,6 +9,7 @@ import {
   OrderedListButton,
   UnderlineButton,
   UnorderedListButton,
+  BlockquoteButton,
 } from 'draft-js-buttons';
 import createEmojiPlugin, { EmojiPluginConfig } from 'draft-js-emoji-plugin';
 import Editor from 'draft-js-plugins-editor';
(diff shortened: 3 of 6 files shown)


src/frontend/scss/objects/_conversation_topic_detail.scss
```diff
@@ -44,6 +44,27 @@
...
+      }
+
+      blockquote:before {
+        color: #ccc;
+        content: '';
+        font-size: 4em;
+        line-height: 0.1em;
+        margin-right: 0.25em;
+        vertical-align: -0.4em;
+      }
```
jbpenrath, line 63: Since, you do not set any content to `:before`, this pseudo-element just add a left blank to the first line of the blockquote. So why not just use `text-indent` to do that ?

src/frontend/scss/objects/_conversation_topic_detail.scss
```diff
@@ -44,6 +44,27 @@
         padding-top: 5px;
         border-top: 1px dashed $gray-200;
       }
+
+      blockquote {
+        background: #f9f9f9;
+        border-left: 10px solid #ccc;
+        margin: 1.5em 10px;
+        padding: 0.5em 10px;
+        quotes: '\201C''\201D''\2018''\2019';
```
jbpenrath, line 53: This attribute is useless here as you did not use `content: open-quote` in `:before`.

jbpenrath (commented):
Really cool to see someone else which sort alphabetically its css attributes !The only best way to have clean css styles 🙃

madmatah (commented):
My comment is somehow related to @jbpenrath comments on the stylesheet.

It would be great to have a similar style between what's displayed in the draft.js editor and the post displayed when we are browsing a topic.

With the current version, a quote is displayed like this in the draft.js editor : 

Screenshot from 2021-01-07 17-50-21

(we can't see any difference with a standard text)

And like this when rendered on a topic detail page : 

Screenshot from 2021-01-07 17-50-29

carofun:
> It would be great to have a similar style between what's displayed in the draft.js editor and the post displayed when we are browsing a topic.

As it is a wiswig it would be actually nice to see what you get...your absolutely right about it !

src/frontend/scss/objects/_conversation_topic_detail.scss
carofun: I've played around and surely didn't clean it...thank you for your feedback

carofun:
The behaviour of the button has been updated so that we can directly see the result in the draftjs editor

Capture d’écran 2021-01-08 à 15 49 23
Capture d’écran 2021-01-08 à 15 49 31

madmatah (approved):
Perfect ! It's good for me but i'm not a Sass expert so @jbpenrath's validation would be a big plus :)

jbpenrath (approved):
👏 !

PR #143, opened 2021-01-27
✨(back) limit forum listing to the current user LTIContext id

madmatah:
## Purpose

When a user is connected to Ashley via LTI, he can see and interact
with all forums he has access to, based on its group membership.

It can be confusing for the user, because he comes from a specific LTI
context and he does not expect to see information related to other
LTI contexts.

In addition to that, group membership synchronization is made during
the LTI authentication process. And it only synchronizes groups for the
current LTI context. So, if we let users access to resources related
to other LTI contexts, we cannot guarantee that user permissions are
correct and it can lead to security issues.

## Proposal

- [x] After LTI authentication, store the current LTI context id in session
- [x] Override django-machina PermissionHandler to hide forums not related to the current LTI context

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
  - fix serialization error in search index update
  - remove deprecated url in favor of re_path 
+ - prevent inactive users from authenticating via LTI
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
  - upgrade django-machina from 1.1.3 to 1.1.4
  - add button to insert quotes in the draftjs editor
+ - limit forum listing to the current user LTIContext id
 
 ### Fixed
 
--- a/src/ashley/auth/backend.py
+++ b/src/ashley/auth/backend.py
@@ -60,7 +60,7 @@ class LTIBackend(ToolboxLTIBackend):
         )
         user_model = get_user_model()
         try:
-            return user_model.objects.get(
+            user = user_model.objects.get(
                 lti_consumer=lti_consumer, lti_remote_user_id=remote_user_id
             )
         except user_model.DoesNotExist:
--- a/src/ashley/machina_extensions/forum_permission/handler.py
+++ b/src/ashley/machina_extensions/forum_permission/handler.py
@@ -2,6 +2,7 @@
 This module defines a ``PermissionHandler`` abstraction that allows to
 implement filter or access logic related to forums.
 """
+from typing import Optional
 
 from machina.apps.forum_permission.handler import (
     PermissionHandler as BasePermissionHandler,
@@ -11,6 +12,28 @@ from machina.apps.forum_permission.handler import (
 class PermissionHandler(BasePermissionHandler):
     """Add ashley specific permission checks to the machina permission handler."""
 
+    def __init__(self):
+        """Constructor"""
+        super().__init__()
+        # This attribute is used to store the LTIContext of the current user.
+        # (see middleware.ForumPermission)
+        # It allows the permission checking functions to be aware of the current LTIContext
+        # we are scoped into, if any.
+        self.current_lti_context_id: Optional[int] = None
+
     def can_rename_forum(self, forum, user):
         """ Given a forum, checks whether the user can rename it. """
         return self._perform_basic_permission_check(forum, user, "can_rename_forum")
+
+    def forum_list_filter(self, qs, user):
+        """
+        Filters the given queryset in order to return a list of forums that
+        can be seen and read by the specified user (at least).
+
+        We override django machina's method to filter forums based on the
+        current LTI context of the User, if any.
+        """
+        forums_to_show = super().forum_list_filter(qs, user)
+        if self.current_lti_context_id:
+            return forums_to_show.filter(lti_contexts__id=self.current_lti_context_id)
+        return forums_to_show
--- a/src/ashley/views.py
+++ b/src/ashley/views.py
@@ -20,6 +20,7 @@ from machina.apps.forum_permission.viewmixins import (
 from machina.core.db.models import get_model
 from machina.core.loading import get_class
 
+from . import SESSION_LTI_CONTEXT_ID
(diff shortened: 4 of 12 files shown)


src/ashley/machina_extensions/forum_permission/handler.py
```diff
@@ -11,6 +12,28 @@
...
     """Add ashley specific permission checks to the machina permission handler."""
 
+    def __init__(self):
+        """Constructor"""
+        super().__init__()
+        # This attribute is used to store the LTIContext of the current user.
+        # (see middleware.ForumPermission)
+        # It allows the permission checking functions to be aware of the current LTIContext
+        # we are scoped into, if any.
+        self.current_lti_context_id: Optional[int] = None
```
sampaccoud, line 22: is it really alway an integer? On all LMS?

src/ashley/machina_extensions/forum_permission/handler.py
```diff
@@ -11,6 +12,28 @@
...
     def can_rename_forum(self, forum, user):
         """ Given a forum, checks whether the user can rename it. """
         return self._perform_basic_permission_check(forum, user, "can_rename_forum")
+
+    def forum_list_filter(self, qs, user):
+        """
+        Filters the given queryset in order to return a list of forums that
+        can be seen and read by the specified user (at least).
+
+        We overrides django machina's method to filter forums based on the
```
sampaccoud, line 33:
```suggestion
        We override django machina's method to filter forums based on the
```

src/ashley/machina_extensions/forum_permission/handler.py
madmatah: This is the primary key of our LTIContext model, so its always an integer.

The "context_id" you are referring to, that is transmitted in the LTI launch request, is actually a string, yes!

sampaccoud: IC :pray:

src/ashley/views.py
```diff
@@ -49,6 +50,9 @@ def _do_on_login(self, lti_request: LTI) -> HttpResponse:
         # Synchronize the user groups related to the current LTI context
         context.sync_user_groups(self.request.user, lti_request.roles)
 
+        # Store the current user LTI context in session
+        self.request.session[SESSION_LTI_CONTEXT_ID] = context.id
```
sampaccoud, line 54: Your are now testing the middleware and the permission handler but I didn't see how you ensure that this line works? Do the tests succeed if you remove it?

madmatah: You're right. This line, or more precisely this class,  is not tested.
I'll add a test for the ForumLTIView.

madmatah: @sampaccoud I added tests for the `ForumLTIView` class.
And at the same time, I discovered/fixed a security issue. Thanks ! :pray:

CHANGELOG.md
```diff
@@ -12,10 +12,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
  - upgrade django-machina from 1.1.3 to 1.1.4
  - add button to insert quotes in the draftjs editor
+ - limit forum listing to the current user LTIContext id
 
 ### Fixed
 
  - fix serialization error in search index update
+ - prevent inactive users from login via LTI
```
sampaccoud, line 20:
```suggestion
 - prevent inactive users from authenticating via LTI
```

src/ashley/auth/backend.py
```diff
@@ -60,7 +60,7 @@ def authenticate(self, request, username=None, password=None, **kwargs):
         )
         user_model = get_user_model()
         try:
-            return user_model.objects.get(
+            user = user_model.objects.get(
```
sampaccoud, line 63: There is a :lock: gitmoji for security fixes.

src/ashley/views.py
sampaccoud: :medal_sports:

tests/ashley/lti_utils.py
```diff
@@ -0,0 +1,47 @@
+"""This package contains helper functions to build LTI requests"""
+
+from urllib.parse import unquote
+
+from oauthlib import oauth1
+
+CONTENT_TYPE = "application/x-www-form-urlencoded"
+
+
+def sign_parameters(passport, lti_parameters, url):
```
sampaccoud, line 10: could also be part of lti toolbox lib in the future...?

tests/ashley/test_forum_lti_view.py
```diff
@@ -0,0 +1,198 @@
...
+        # Our current LTIContext id should have been injected in the user session
+        self.assertEqual(context.id, self.client.session.get(SESSION_LTI_CONTEXT_ID))
+
+        # The launch_presentation_locale should be set in the language cookie
+        self.assertEqual(
+            "en", response.client.cookies[settings.LANGUAGE_COOKIE_NAME].value
+        )
+
+    def test_login_inactive_user(self):
+        """Test LTI launch URL view with an inactive user"""
```
sampaccoud, line 76:
```suggestion
        """An inactive user should not be allowed to authenticate via a valid LTI request"""
```

tests/ashley/test_forum_lti_view.py
```diff
@@ -0,0 +1,198 @@
...
+class ForumLTIViewTestCase(TestCase):
+    """Test the ForumLTIView class"""
+
+    def setUp(self):
+        """Override the setUp method to instanciate and serve a request factory."""
+        super().setUp()
+        self.request_factory = RequestFactory()
+        self._auth_backend = LTIBackend()
+
+    def test_login_success(self):
```
sampaccoud, line 29: I noticed your test methods do not follow our handbook :point_up:

madmatah: I updated the method names and docstring.
But if your comment was about something else that I missed, please tell me :pray:

tests/ashley/lti_utils.py
madmatah: In fact, it is part of lti toolbox : https://github.com/openfun/django-lti-toolbox/blob/master/tests/lti_toolbox/utils.py

But since it's in the `tests` namespace, it's not bundled with the package on pypi.
I'll take it if you have good practices to share on this subject.

sampaccoud (approved):

PR #144, opened 2021-01-28
✨(front) add an icon for instructor on topic and post

carofun:
## Purpose

We want to identify when a topic or a post has been created by an instructor

## Proposal

We add an icon next to the username of an instructor when it's the case.

Capture d’écran 2021-02-02 à 13 42 16
Capture d’écran 2021-02-02 à 13 42 09
Capture d’écran 2021-02-02 à 13 41 57

### DIFF
--- a/src/ashley/models.py
+++ b/src/ashley/models.py
@@ -99,6 +99,13 @@ class AbstractLTIContext(Model):
         """
         return self.GROUP_DELIMITER.join([self.GROUP_PREFIX, str(self.id)])
 
+    def get_group_role_name(self, role) -> str:
+        """
+        Get the name of the group for a given role
+        By default it is of the form "cg:{context.id}:role:{role}"
+        """
+        return self.GROUP_DELIMITER.join([self.base_group_name, "role", role])
+
     def get_role_groups(self, roles: List[str]) -> List[Group]:
         """
         Get the Django Groups corresponding to a list of LTI Context Roles
@@ -115,7 +122,7 @@ class AbstractLTIContext(Model):
         All LTI user authenticated within this LTI context and having this role
         should be in this group.
         """
-        group_name = self.GROUP_DELIMITER.join([self.base_group_name, "role", role])
+        group_name = self.get_group_role_name(role)
         return self._get_or_create_group(group_name)
 
     @staticmethod
@@ -150,6 +157,18 @@ class AbstractLTIContext(Model):
             if group not in current_groups:
                 logger.debug("Add user %s to group %s", user, group)
                 user.groups.add(group)
+
+    def get_user_roles(self, user) -> List[str]:
+        """
+        Return list of roles name without group pattern
+        cg:{context.id}:role: for this user in this LTI Context
+        """
+        role_group_prefix = self.get_group_role_name("")
+        user_role_groups = user.groups.filter(
+            name__startswith=role_group_prefix
+        ).values_list("name", flat=True)
+        roles = [group.replace(role_group_prefix, "") for group in user_role_groups]
+        return roles
 
     class Meta:
         """Options for the ``AbstractLTIContext`` model."""
--- a/src/ashley/templates/forum/forum_list.html
+++ b/src/ashley/templates/forum/forum_list.html
@@ -77,6 +77,7 @@
         {% blocktrans trimmed with poster_url=poster_url username=node.last_post.poster|forum_member_display_name %}
         By: <a href="{{ poster_url }}">{{ username }}</a>
         {% endblocktrans %}
+        {% include "partials/user_icon.html" with topic=node.last_post.topic user=node.last_post.poster %}
         {% else %}
         {% blocktrans trimmed with poster_username=node.last_post.username %}
         By: {{ poster_username }}
@@ -150,6 +151,7 @@
         {% blocktrans trimmed with poster_url=poster_url username=node.last_post.poster|forum_member_display_name %}
         By: <a href="{{ poster_url }}">{{ username }}</a>
         {% endblocktrans %}
+        {% include "partials/user_icon.html" with topic=node.last_post.topic user=node.last_post.poster %}
         {% else %}
         {% blocktrans trimmed with poster_username=node.last_post.username %}
         By: {{ poster_username }}
--- a/src/ashley/templates/forum_conversation/topic_detail.html
+++ b/src/ashley/templates/forum_conversation/topic_detail.html
@@ -1,0 +1,121 @@
+{% extends 'board_base.html' %}
+{% load i18n %}
+{% load forum_conversation_tags %}
+{% load forum_member_tags %}
+{% load forum_permission_tags %}
+
+{% block sub_title %}{{ topic.subject }}{% endblock sub_title %}
+
+{% block content %}
+<div class="row"><div class="col-12"><h1>{{ topic.subject }}</h1></div></div>
+<div class="row">
+  <div class="col-12 col-md-4 topic-actions-block">
+  {% include "forum_conversation/partials/topic_detail_actions.html" %}
+  </div>
+  <div class="col-12 col-md-8 pagination-block">
+  {% with "pagination-sm justify-content-end" as pagination_size %}
+  {% include "partials/pagination.html" %}
+  {% endwith %}
+  </div>
+</div>
+<div class="row">
+  <div class="col-12">
+  {% for post in posts %}
+    {% if forloop.first and post.is_topic_head and poll %}
+    {% include "forum_conversation/forum_polls/poll_detail.html" %}
+    {% endif %}
+    <div id="{{ post.id }}" class="my-3 card post">
+      <div class="card-body">
+        <div class="row">
+          <div class="col-md-10 post-content-wrapper">
+            <div class="float-right post-controls">
+              {% get_permission 'can_edit_post' post request.user as user_can_edit_post %}
+              {% if user_can_edit_post %}
+              <a href="{% if post.is_topic_head %}{% url 'forum_conversation:topic_update' forum.slug forum.pk topic.slug topic.pk %}{% else %}{% url 'forum_conversation:post_update' forum.slug forum.pk topic.slug topic.pk post.pk %}{% endif %}" class="btn btn-warning btn-sm" title="{% trans "Edit" %}"><i class="fas fa-edit"></i>&nbsp;{% trans "Edit" %}</a>
+              {% endif %}
+              {% get_permission 'can_delete_post' post request.user as user_can_delete_post %}
+              {% if user_can_delete_post %}
+              <a href="{% url 'forum_conversation:post_delete' forum.slug forum.pk topic.slug topic.pk post.pk %}" class="btn btn-danger btn-sm" title="{% trans "Delete" %}"><i class="fas fa-times"></i></a>
+              {% endif %}
+            </div>
+              {% spaceless %}
+              <h4 class="m-0 subject">
+                {{ post.subject }} 
+                &nbsp;<a href="{% url 'forum_conversation:topic' forum.slug forum.pk topic.slug topic.pk %}?post={{ post.pk }}#{{ post.pk }}">&#182;</a>
+              </h4>
+              {% endspaceless %}
+              <p>
+                <small class="text-muted">
+                {% spaceless %}
+                <i class="fas fa-clock"></i>&nbsp;
+                {% if post.poster %}
+                {% url 'forum_member:profile' post.poster_id as poster_url %}
+                {% blocktrans trimmed with poster_url=poster_url username=post.poster|forum_member_display_name%}
+                  By: <a href="{{ poster_url }}">{{ username }}</a>
+                {% endblocktrans %}
+                {% include "partials/user_icon.html" with topic=post.topic user=post.poster %}
+                {% blocktrans trimmed with creation_date=post.created %}
+                on {{ creation_date }}
+                {% endblocktrans %}
+                {% else %}
+                {% blocktrans trimmed with poster_username=post.username creation_date=post.created %}
+                  By: {{ poster_username }} on {{ creation_date }}
+                {% endblocktrans %}
+                {% endif %}
+                {% endspaceless %}
+                </small>
+              </p>
+              <div class="post-content">
+                {{ post.content.rendered }}
+              </div>
+              {% include "forum_conversation/forum_attachments/attachments_detail.html" %}
+              {% if post.enable_signature and post.poster.forum_profile.signature %}
+              <div class="post-signature">
+                {{ post.poster.forum_profile.signature.rendered }}
+              </div>
+              {% endif %}
+              {% if post.updates_count %}
+              <div class="mt-4 edit-info">
+                <small class="text-muted">
+                  <i class="fas fa-edit"></i>&nbsp;{% if post.updated_by %}{% trans "Last edited by:" %}&nbsp;<a href="{% url 'forum_member:profile' post.updated_by_id %}">{{ post.updated_by|forum_member_display_name }}</a>&nbsp;{% else %}{% trans "Updated" %}&nbsp;{% endif %}{% trans "on" %}&nbsp;{{ post.updated }}, {% blocktrans count counter=post.updates_count %}edited {{counter }} time in total.{% plural %}edited {{counter }} times in total.{% endblocktrans %}
+                </small>
+                {% if post.update_reason %}
+                <br />
+                <small class="text-muted">
+                  <b>{% trans "Reason:" %}</b>&nbsp;{{ post.update_reason }}
+                </small>
+                {% endif %}
+              </div>
+              {% endif %}
+          </div>
+          <div class="col-md-2 d-none d-md-block post-sidebar">
+            {% if post.poster %}
+            <div class="avatar">
+              <a href="{% url 'forum_member:profile' post.poster_id %}">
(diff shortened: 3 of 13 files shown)


src/ashley/models.py
```diff
@@ -151,6 +152,15 @@ def sync_user_groups(self, user, roles: List[str]) -> None:
...
                 user.groups.add(group)
 
+    def is_user_instructor(self, user) -> bool:
+        """
+        Return if the user is part of group instructor for this LTI Context
+        """
+        return (
+            user.groups.filter(name=self.get_role_group(self.GROUP_INSTRUCTOR)).count()
+            > 0
+        )
```
sampaccoud, line 162:
```suggestion
        return user.groups.filter(name__endswith=f"role:{INSTRUCTOR_GROUP}").exists()
```

src/ashley/models.py
```diff
@@ -151,6 +152,15 @@ def sync_user_groups(self, user, roles: List[str]) -> None:
                 logger.debug("Add user %s to group %s", user, group)
                 user.groups.add(group)
 
+    def is_user_instructor(self, user) -> bool:
```
madmatah, line 155: I'm not confortable with such a specific method in the `LTIContext` model, as it adds the interpretation of the LTI roles to this model, in addition to reflecting their composition.

What do you think about adding a method `get_user_roles(self, user) -> List[str]` here, and handle the specific detection of the "instructor" role outside of this model ?

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Integration tests to validate that the filter is_user_instructor
+    can detect if user is instructor of a forum
+    """
+
+    def setUp(self):
+        """
+        Set up needed data for our tests
+        """
+        # Create two LTI Context
```
sampaccoud, line 19: I wouldn't put this in the setUp. It's a lot of things you do here and you will later add more tests, which will not all need exactly this setup... but it will be created nevertheless and slow down tests considerably.
My personal practice is to have a specific method (if the name does not start by test_ it will not be treated as a test by the runner. Make it start by _ for example) that I explicitly call from everytest that needs exactly this setup.
At this stage, you could also just put the code in your test since there's only 1 :wink:

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -0,0 +1,68 @@
...
+        self.forum.lti_contexts.add(context1)
+
+        # Set up topic
+        self.topic = TopicFactory(
+            forum=self.forum, poster=self.user1, subject="topic création"
+        )
+
+    def test_can_tell_if_the_user_is_instructor_of_this_forum(self):
+        """
+        Givig two users. User1 is student of the forum, User2 is instructor.
```
sampaccoud, line 43:
```suggestion
        Given two users. User1 is student of the forum, User2 is instructor.
```

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -0,0 +1,68 @@
...
+        We control that is_user_instructor can detect if a user is an instructor.
+        Then a forum can be part of multiple context. If a user is instructor in one context,
```
sampaccoud, line 45:
```suggestion
        Then a forum can be part of multiple contexts. If a user is instructor in one context,
```

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -0,0 +1,68 @@
...
+        he is considered admin of this forum in all context. We add a new context to the forum
```
sampaccoud, line 46:
```suggestion
        he is considered admin of this forum in all contexts. We add a new context to the forum
```

src/ashley/templatetags/custom_tags.py
```diff
@@ -0,0 +1,39 @@
...
+@register.filter()
+def is_user_instructor(topic_pk, user):
+    """
+    This will return a boolean indicating if the passed user is instructor
+    for the given topic.
+    Usage::
+        {% if topic_pk|is_user_instructor:user %}...{% endif %}
+    """
+    # get the list of lti concerns by this forum, one forum can have multiple LTI Context
+    # if the user is instructor in one of them then he is considered
```
sampaccoud, line 22: I'm confused: do we still want this now that @madmatah's PR https://github.com/openfun/ashley/pull/143 is limiting a user to a context for a given LTI connection?

carofun: This aspect will disappear but as it is still for now the attended behaviour, I think until it evolved I have to test this.

tests/ashley/test_templatetags_custom_tags.py
carofun: ok thank you for the best practice

src/ashley/templates/forum_conversation/topic_detail.html
```diff
@@ -0,0 +1,121 @@
+{% extends 'board_base.html' %}
```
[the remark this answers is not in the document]
carofun: I agree with you I had trouble at first to see this code that seems to be repeated. That would be a nice next step :-) I hope I can still validate this PR first :-)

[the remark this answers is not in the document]
sampaccoud: Well this PR is 99% ready to merge imo but if you think you can get these blocks in a release in a few days maybe you should wait for it? Otherwise, we can just open an issue and list all the templates that we want to stop overriding in full (I think I remember seeing other cases in Ashley already). Then the PR to django-machina could cover all templates :taco:

src/ashley/templatetags/custom_tags.py
sampaccoud: I'm afraid he will merge his before yours ^^

carofun: I'm not sure regarding your return, I have a good chance !

src/ashley/templates/forum_conversation/topic_detail.html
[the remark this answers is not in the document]
carofun: Either way I would have to rewrite some code and I have no influence regarding when this work could be accepted in django machina. I would propose :-) for both project to advance to validate this PR in our project, submit the work in machina and when this is done to do a new PR on our project. What do you think of this strategy ?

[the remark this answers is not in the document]
sampaccoud: Yes :+1: let's open an issue in Ashley listing all templates that we have overriden and would want to modify in django-machina.

src/ashley/templatetags/custom_tags.py
```diff
@@ -0,0 +1,29 @@
...
+    Usage::
+        {% if topic|is_user_instructor:user %}...{% endif %}
+    """
+    # get the list of lti concerns by this forum, one forum can have multiple LTI Context
+    # if the user is instructor in one of them then he is considered
+    # as instructor for this forum
+    return any(
+        _FORUM_ROLE_INSTRUCTOR in lti.get_user_roles(user)
+        for lti in LTIContext.objects.filter(
+            forum__id=Topic.objects.get(pk=topic.pk).forum.id
```
sampaccoud, line 27:
```suggestion
            forum=topic.forum
```

src/ashley/templates/forum/forum_list.html
```diff
@@ -77,6 +77,7 @@ <h3 class="m-0 card-title h5 text-dark">{% trans "Forums" %}</h3>
         {% blocktrans trimmed with poster_url=poster_url username=node.last_post.poster|forum_member_display_name %}
         By: <a href="{{ poster_url }}">{{ username }}</a>
         {% endblocktrans %}
+        {% include "partials/user_icon.html" with topic=node.last_post.topic user=node.last_post.poster %}
```
sampaccoud, line 80: I would have added a minimal integration test that renders the forum list and confirms the presence of the icon 
Same for other places where you add the icon :man_shrugging:

sampaccoud (approved):

src/ashley/templates/forum/forum_list.html
carofun: ok :-) I had them on the first version because of the way I had build it and was frustrated that it wasn't checked anymore that was probably the reason why, I still had to do them :-)

src/ashley/models.py
```diff
@@ -151,6 +157,19 @@ def sync_user_groups(self, user, roles: List[str]) -> None:
...
+        """
+        Return list of roles name without group pattern
+        cg:{context.id}:role: for this user in this LTI Context
+        """
+        role_group_prefix = self.get_group_role_name("")
+        user_role_groups = user.groups.filter(
+            name__startswith=role_group_prefix
+        ).values_list("name", flat=True)
+        roles = [group.replace(role_group_prefix, "") for group in user_role_groups]
+        logger.debug("List of roles %s", roles)
```
madmatah, line 170: If you want to keep this debug, I suggest you to include the user in the logged string, otherwise it's a difficult to do something useful with this information.

src/ashley/templates/forum/forum_list.html
carofun: Sam I've added tests for this purpose, I've resubmitted a new review for this. thanks in advance

src/ashley/models.py
```diff
@@ -99,6 +99,12 @@ def base_group_name(self) -> str:
         """
         return self.GROUP_DELIMITER.join([self.GROUP_PREFIX, str(self.id)])
 
+    def get_group_role_name(self, role) -> str:
+        """
+        Get the name of the group for a given role
```
sampaccoud, line 104:
```suggestion
        Get the name of the group for a given role
        By default it is of the form "cg:{context.id}:role:{role}"
```

src/ashley/templates/partials/user_icon.html
```diff
@@ -0,0 +1,6 @@
+{% load i18n %}
+{% load custom_tags %}
+
+{% if topic|is_user_instructor:user %}
+    <i class="icon_writer fas fa-award" aria-hidden="true" title="{%trans 'Instructor'%}"></i>
```
sampaccoud, line 5: I'm not sure about accessibility. You're hiding it but it conveys some important information. WDYT @mbenadda?

src/ashley/templatetags/custom_tags.py
```diff
@@ -0,0 +1,27 @@
...
+    for the given topic.
+    Usage::
+        {% if topic|is_user_instructor:user %}...{% endif %}
+    """
+    # get the list of lti concerns by this forum, one forum can have multiple LTI Context
+    # if the user is instructor in one of them then he is considered
+    # as instructor for this forum
+    return any(
+        _FORUM_ROLE_INSTRUCTOR in lti.get_user_roles(user)
+        for lti in LTIContext.objects.filter(forum=Topic.objects.get(pk=topic.pk).forum)
```
sampaccoud, line 26: You insist making a get query? Is it not working with the topic you get from args?
```suggestion
        for lti in LTIContext.objects.filter(forum=topic.forum)
```

tests/ashley/test_models_lticontext.py
```diff
@@ -112,3 +112,50 @@ def test_sync_user_groups(self):
             ],
             list(user.groups.values_list("name", flat=True)),
         )
+
+    def test_get_group_role_name(self):
+        """Test get_group_role_name that returns the name of the role with the specific pattern"""
```
sampaccoud, line 117: I would say we usually use imperative form for test docstrings (maybe it could go to the handle book if everyone agrees with this):
```suggestion
        """get_group_role_name should return the name of the role with the specific pattern"""
```

tests/ashley/test_models_lticontext.py
```diff
@@ -112,3 +112,50 @@ def test_sync_user_groups(self):
...
+        # Ensure that the get_group_role_name generator respect a specific name pattern
+        self.assertEqual(f"cg:{context.id}:role:", context.get_group_role_name(""))
+        # Test with a actual group name label
+        name_instructor_group = f"cg:{context.id}:role:instructor"
+        self.assertEqual(
+            name_instructor_group, context.get_group_role_name("instructor")
+        )
+
+    def test_get_user_roles(self):
+        """Test get_user_roles that returns the list of the roles names
```
sampaccoud, line 129:
```suggestion
        """get_user_roles should return the list of the roles names``

File: tests/ashley/test_templatetags_custom_tags.py [hunk 17]
```diff
@@ -0,0 +1,265 @@
...
+            poster=user1,
+            subject="post by instructor",
+        )
+
+        self.client.force_login(user1)
+
+        # accessing forum view
+        response = self.client.get(reverse("forum:index"))
+        # check that it contains the icon instructor
+        self.assertIn(
```
sampaccoud, line 180: you should be able to use
```
self.assertContains(response, rendered_content, html=True) 
```
Which takes care of trimming and utf8 encoding.

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def test_render_block_is_instructor_of_this_forum(self):
+        """
+        check that the instructor icon is present on the different views
+        as expected
+        """
+        created_by = 'By: <a href="{{url}}">{{ username }}</a>'
+        icon = '<i class="icon_writer fas fa-award" aria-hidden="true" title="Instructor"></i>'
+
+        def get_rendered(topic, user, selector=""):
```
sampaccoud, line 123: I would write the expected html directly in the test instead of going through template rendering because this is what you are trying to test. If your template for some reason renders "hey", then you will test that "hey" is also in your response... so the test will pass but the feature is broken.

src/ashley/templatetags/custom_tags.py
carofun: ah ouups it's magic !

tests/ashley/test_templatetags_custom_tags.py
carofun: ah ah ah...

src/ashley/templates/partials/user_icon.html
carofun: Following the instructions of font awesome I think we're good there, optionally they recommend to add a span to describe more, but in our case I think this is sufficient. Machina is not really good at it, for instance it as no title on href of the same block By: <a href="{{ poster_url }}">{{ username }}</a>

ref https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility

Web Fonts with CSS: Semantic Icons 

If your icons have semantic meaning, you’ll need to manually add a few things so that your icon is appropriately accessible:

aria-hidden="true" attribute.
Provide a text alternative inside a <span> (or similar) element. Also include appropriate CSS to visually hide the element while keeping it accessible to assisitive technologies.
title attribute on the icon to provide a tooltip for sighted mouse users.

sampaccoud: So if I understand well we are missing this part?

> Provide a text alternative inside a (or similar) element. Also include appropriate CSS to visually hide the element while keeping it accessible to assisitive technologies.

sampaccoud: My understanding is that if we don't add this alternative text, screen readers will not see anything to indicate that the user is an instructor and they will therefore have less information than other users which is not what we want.

tests/ashley/test_templatetags_custom_tags.py
carofun: Thank you to have share this option with me, I hadn't seen it yesterday. I'm sure in the futur it will be of good use.
Unfortunately I've tried again with this solution and I can't manage to find what I'm looking for, I think my problem remains on "\n". As my previous solution is working and as I've already spent a good bit of time on it yesterday, I submit again with no change on this point.
```
        By: <a href="/forum/member/profile/3/">instructor</a>




    <i class="icon_writer fas fa-award" aria-hidden="true" title="Instructor"></i>
```

tests/ashley/test_templatetags_custom_tags.py
carofun: they were no point on using Template the way I did it, I've replaced it with a f" string.

src/ashley/templates/partials/user_icon.html
carofun: I understand that if we want to do the top solution yes but it's out of context IMO, it's too much for too little.
But by adding the title we already provide a tooltip for sighted mouse users which is a good first step.

mbenadda: The current implementation is not accessible unless the icon is purely decorative and conveys 0 meaning.

If there's a meaning we need to add a visually hidden (`sr-only` or whatever the helper class is called in Ashley) span with the same text as the one in the `<i>` title.

I think font-awesome themselves recommend this approach for non interactive icons that are supposed to carry meaning.

carofun: yes they do indeed as mentioned above, I doubt though the need of this in this context but I will do so :-)

mbenadda: I did not work on Ashley so I'm not as knowledgeable as you about the context here.

I think in general though we should work from the assumption that if any user could see this UI element, then it needs to be accessible. Apart from my personal opinion, it's also necessary if we aim to be RGAA compliant and hopefully pass an audit later on.

carofun: I think you are absolutely right and I got lazy for no reason on this one ! thank you for you feedback.

PR #147, opened 2021-02-05
🎨(test) implement simple html content tests

carofun:
## Purpose

Testing the render of html for the reliability of the project
is essential. Writing these tests in the most simpliest way
is a key point on expanding and maintaining them. We use an
helper dependencie django-with-asserts to help us on redacting
them.

## Proposal

Adding the dependencie django-with-asserts and rewrite the test that target specific element in html

### DIFF
--- a/tests/ashley/test_templatetags_custom_tags.py
+++ b/tests/ashley/test_templatetags_custom_tags.py
@@ -1,13 +1,15 @@
 """Test suite for ashley custom tags"""
 import logging
-import re
 from typing import List
 
 from django.template import Context, Template
 from django.test import TestCase
+from django.test.html import parse_html
 from django.urls import reverse
+from lxml import etree
 from machina.apps.forum_permission.shortcuts import assign_perm
 from machina.core.loading import get_class
+from with_asserts.mixin import AssertHTMLMixin
 
 from ashley.defaults import (
     DEFAULT_FORUM_BASE_PERMISSIONS,
@@ -29,7 +31,7 @@ get_forum_member_display_name = get_class(
 logger = logging.getLogger(__name__)
 
 
-class TestIsUserInstructorTag(TestCase):
+class TestIsUserInstructorTag(TestCase, AssertHTMLMixin):
     """
     Integration tests to validate that the filter is_user_instructor
     can detect if user is instructor of a forum
@@ -57,11 +59,6 @@ class TestIsUserInstructorTag(TestCase):
         """Grant a list of permissions for a group on a specific forum."""
         for perm in permissions:
             assign_perm(perm, group, forum, True)
-
-    @staticmethod
-    def _trim(answer):
-        """Get rid of space and \n to be able to compare the render of template"""
-        return re.sub(r"\s+", "", answer)
 
     @staticmethod
     def test_can_tell_if_the_user_is_instructor_of_this_forum():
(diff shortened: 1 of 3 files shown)


tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -202,26 +220,46 @@ def get_html_post_by(user):
...
-        )
-        # check that the last post creation writer doesn't contain the icon for user student
-        self.assertNotIn(
-            get_html_post_by(user2),
-            self._trim(response.content.decode("utf-8")),
-        )
+        with self.assertHTML(response, ".topic-last-post") as (html,):
+            extract = parse_html(str(etree.tostring(html)))
+            # control that the instructor's icon is present
+            self.assertFalse(
```
sampaccoud, line 245: when you test that something is NOT in the response, it's better to  test the minimum. If "instructor" in extract is False then it's a better test than testing that this whole sentence is not exactly in the response...

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -160,36 +136,78 @@ def get_html_post_by(user):
         self._init_forum(forum, context)
 
         # Set up topic
-        topic1 = TopicFactory(forum=forum, poster=user1, subject="topic by instructor")
+        topic1 = TopicFactory(
+            forum=forum, poster=user1, subject="Je ne suis pas votre élève"
+        )
```
sampaccoud, line 141: Actually you don't need to worry about the value of the subject string since this is not the subject of this test...
```suggestion
        topic1 = TopicFactory(forum=forum)
```

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -160,36 +136,78 @@ def get_html_post_by(user):
...
 
         # Set up post
         PostFactory.create(
             topic=topic1,
             poster=user1,
-            subject="post by instructor",
+            subject="Vous n'êtes pas mon professeur",
```
sampaccoud, line 147: ditto
```suggestion
```

tests/ashley/test_templatetags_custom_tags.py
```diff
@@ -160,36 +136,78 @@ def get_html_post_by(user):
...
         )
 
+        # log user1
         self.client.force_login(user1)
 
         # accessing forum view
         response = self.client.get(reverse("forum:index"))
+        with self.assertHTML(response, ".forum-last-post") as (html,):
+            extract = parse_html(str(etree.tostring(html)))
```
sampaccoud, line 156: I found a better name than `extract` :
```suggestion
            fragment = parse_html(str(etree.tostring(html)))
```

sampaccoud (approved):

tests/ashley/test_templatetags_custom_tags.py
carofun: ok I kept the relation with the poster though !

tests/ashley/test_templatetags_custom_tags.py
carofun: that's a good point indeed !

madmatah (approved):
I'm scared by the version number of `django-with-asserts` and the last commit date (2013).

BUT,

carofun:
ah ah I love your video
django-with-asserts is not of much used in reality it only do a basic use of lxml
If we keep on using this for our html test, maybe we were saying with Sam we could use something of our own that does the same job and gets extended if we need to

> I'm scared by the version number of `django-with-asserts` and the last commit date (2013).
> 
> BUT,
>

PR #150, opened 2021-02-09
🗑(sandbox) remove deprecated url in favor of re_path

carofun:
## Purpose

Django 3.1 version have introduced deprecated method
RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor
of django.urls.re_path()

## Proposal

Update the code to follow new recommendations and replace  django.conf.urls.url() by django.urls.re_path()

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
  - fix serialization error in search index update
+ - remove deprecated url in favor of re_path 
 
 ### Added
 
--- a/sandbox/dev_tools/urls.py
+++ b/sandbox/dev_tools/urls.py
@@ -1,18 +1,17 @@
 """URLs for the dev_tools django application."""
-from django.conf.urls import url
-from django.urls import reverse_lazy
+from django.urls import re_path, reverse_lazy
 from django.views.generic import RedirectView
 
 from . import views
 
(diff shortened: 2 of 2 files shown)


sandbox/urls.py
```diff
@@ -5,13 +5,13 @@
...
-from django.urls import include, path
+from django.urls import include, re_path
 
 from ashley import urls as ashley_urls
 
 urlpatterns = [
-    path("admin/", admin.site.urls),
-    path("", include(ashley_urls)),
+    re_path("admin/", admin.site.urls),
+    re_path("", include(ashley_urls)),
```
madmatah, line 14: `path()` is not deprecated and we don't need regex here, why changing this ?

sandbox/urls.py
```diff
@@ -5,13 +5,13 @@
...
 ]
 if DevToolsAppConfig.name in settings.INSTALLED_APPS:
-    urlpatterns += [path("dev/", include(dev_urls))]
```
madmatah, line 17: ditto

sampaccoud:
@carofun note that you have a new "path" item for urls that don't need to extract variables from the path. You can mix `path` and `re_path` depending on the case e.g https://github.com/openfun/richie/blob/master/sandbox/urls.py#L27.

madmatah:
Thank you for this PR @carofun ! :pray:

madmatah (approved):

PR #151, opened 2021-02-11
🚨(linter) fix mypy target directory

carofun:
## Purpose

In our developpement environnement we have an error on the make lint command.

## Proposal

mypyp is actually checking the folder frontend and generating useless error in our make lint command.
We configure it so that it checks only the right files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,7 +136,7 @@ jobs:
           command: ~/.local/bin/bandit -qr src/ashley sandbox
       - run:
           name: Type-check code with mypy
-          command: ~/.local/bin/mypy src
+          command: ~/.local/bin/mypy src/ashley
 
   test-back:
     docker:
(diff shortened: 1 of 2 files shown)


madmatah (commented):
Thanks for this fix!

A few comments : 

1. `mypy` is also used in CircleCI configuration, I suggest to update it to check the same directories that the `make lint-mypy` command.

2. About the commit message title, I suggest the following : 
>    :rotating_light: (linter) fix mypy target directory

   (because it was not a warning but an error, and the `node_modules` and `mypy` words side by side can be confusing)

3. In the commit message body :
 - typo : "`mypy`" instead of "mypyp"
 - "the `src/frontend` folder " instead of "the folder frontend"
 - "in our `make lint-mypy`" instead of "in our make lint" to be more specific

madmatah (commented):
1 typo remaining in the commit body : `lint-mpy` => `lint-mypy`.

madmatah (approved):

PR #152, opened 2021-02-12
(models) enforce integrity between User and LTIContext

carofun:
## Purpose

Currently with the method sync_user_group we can add groups to the user even if this one is not part of the lticontext. This shouldn't be allowed. 

## Proposal

We control that user is part of the lticontext before adding groups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - fix serialization error in search index update
  - remove deprecated url in favor of re_path 
  - prevent inactive users from authenticating via LTI
+ - prevent linking a User to a LTIContext related to different LTI Consumer
 
 ### Added
 
--- a/src/ashley/models.py
+++ b/src/ashley/models.py
@@ -4,6 +4,7 @@ from typing import List
 
 from django.contrib.auth.models import AbstractUser as DjangoAbstractUser
 from django.contrib.auth.models import Group
+from django.core.exceptions import PermissionDenied
 from django.db import models
 from django.db.models import Model
 from django.utils.translation import gettext_lazy as _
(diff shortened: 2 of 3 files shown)


src/ashley/models.py
```diff
@@ -139,6 +140,9 @@ def sync_user_groups(self, user, roles: List[str]) -> None:
         """
         Synchronize the group membership of a user for this LTI Context
         """
+        # Control user is part of this lti_context
+        if user.lti_consumer_id != self.lti_consumer_id:
+            raise PermissionDenied(_("user must be part of this lti context"))
```
madmatah, line 145: In fact, we call this function to link a user to a specific LTI context.
I suggest the following error instead : 

```suggestion
            raise PermissionDenied(_("The User and LTIContext must be part of the same LTI Consumer"))
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - fix serialization error in search index update
  - remove deprecated url in favor of re_path 
  - prevent inactive users from authenticating via LTI
+ - check user is part of lticontext before adding groups
```
madmatah, line 23:
```suggestion
 - prevent linking a User to a LTIContext related to different LTI Consumer
```

(see my comment on models.py).

madmatah (commented):
I think there is a mixup between LTIContext and LTIConsumer in your commit message.

src/ashley/models.py
```diff
@@ -139,6 +140,11 @@ def sync_user_groups(self, user, roles: List[str]) -> None:
         """
         Synchronize the group membership of a user for this LTI Context
         """
+        # Control user is part of this lti_context
+        if user.lti_consumer_id != self.lti_consumer_id:
+            raise PermissionDenied(
+                _("The User and LTIContext must be part of the same LTI Consumer")
+            )
```
madmatah, line 147:
```suggestion
            raise PermissionDenied("The User and LTIContext must be part of the same LTI Consumer")
```

You don't need to translate exceptions.

madmatah (approved):

PR #156, opened 2021-02-17
⚡️(back) optimize query on accessing forum from LTI launch URL

carofun:
## Purpose

Previous commit `(back) allow identical LTI launch URL for two distinct forums`
removed  the unique property of the field `lti_id`. As a result, the Index
automatically created with the unique property got removed. This field is used
by the LTI launch URL and it’s essential in terms of performance that we
maintain an Index property on this field. 

## Proposal

Add explicitly an Index on`lti_id`  field.


madmatah:
A few suggestions about the commit message :

1)  I suggest to use commit hash to refer to a previous commit, it's more specific, easier to search and github generates links automatically :

    > In commit f8943b5a1d4ccd8823fd204791b4f7a2d5343277, we removed the unique property of the field lti_id.  [...] 

2)  > This field is used by the LTI launch URL

    A field used by a URL seems strange to me.
    Suggestion: «This field is used by the `ForumLTIView` to find the Forum »

3) About the commit message subject and Changelog entry, what do you think about «add index to the Forum model» ?
    Your commit message body explains the context and this subject explains simply the change.

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - limit forum listing to the current user LTIContext id
  - add code blocks in the editor and auto-highlight code in the forum
  - create a forum for each LTIContext id
-
+ - add an index on `lti_id` field in forum table
```
madmatah, line 18:
```suggestion
 - add an index on `lti_id` field in Forum model
```

madmatah (approved):

PR #159, opened 2021-02-23
✨(forum) add mentions in editor for active users in current topic

carofun:
## Purpose

A nice feature in a forum is to be able to mention someone specific using the `@` keyword. It enables to target a person directly and to fluidly exchanges. We’ve added a draft-js plugin to do this task and propose to mention any user that has been active in the current topic.
This is a first step on notifying users.  We can imagine later on exploiting these datas to send external notifications or to centralise them in a user's view.

## Proposal

Add a draft-js plugin to mention users and connect it with Ashley.

Capture d’écran 2021-02-23 à 10 44 09

Capture d’écran 2021-02-23 à 10 38 10
Capture d’écran 2021-02-23 à 10 40 17
Capture d’écran 2021-02-23 à 10 40 27
Capture d’écran 2021-02-23 à 10 41 14


madmatah (changes requested):
Thanks @carofun for this great feature!
My first comment is only based on your screenshots.

I see that you display the `username` field of the User model (e.g. `jojo@dev-consumer`).
This is an internal identifier that should not be displayed to users.

Django-machina provides a method to get a public username (`apps.forum_member.shortcuts.get_forum_member_display_name(user)`) that is configurable with a setting (`MACHINA_USER_DISPLAY_NAME_METHOD`). You should use it instead.

carofun:
> Thanks @carofun for this great feature!
> My first comment is only based on your screenshots.
> 
> I see that you display the `username` field of the User model (e.g. `jojo@dev-consumer`).
> This is an internal identifier that should not be displayed to users.
> 
> Django-machina provides a method to get a public username (`apps.forum_member.shortcuts.get_forum_member_display_name(user)`) that is configurable with a setting (`MACHINA_USER_DISPLAY_NAME_METHOD`). You should use it instead.

I did a fixup regarding this point. Thank you 🙏

src/ashley/machina_extensions/forum_conversation/forms.py
```diff
@@ -0,0 +1,55 @@
...
+                )
+                if User.objects.get(pk=user["poster"])
+                else user["username"],
+                "link": reverse("forum_member:profile", kwargs={"pk": user["poster"]}),
+            }
+            for user in list(active_post_users)
+        ]
+        self.fields["content"].widget.attrs["mention_users"] = sorted(
+            list_active_users, key=lambda i: i["name"]
+        )
```
sampaccoud, line 51: - I noticed that you are ordering twice, so ordering the first query is useless
- You are filtering the first query on active users so you are missing the anonymous users that only keyed in their username without being authenticated (https://github.com/ellmetha/django-machina/blob/master/machina/apps/forum_conversation/abstract_models.py#L245). 
- I think you should probably handle anonymous users and authenticated users in separate queries. For authenticated users, a query on the User model will be more natural. For anonymous users, the "link" property will by null because they don't have a profile right?

sampaccoud: For anonymous users:
```
anonymous_posters = (
    self.topic.posts.filter(approved=True, poster__is_null=True, username__is_null=False)
    .values("username", flat=True)
)
```

For authenticated users:
```
authenticated_users = (
    User.objects.filter(is_active=True).filter(posts__topic=self.topic, posts__approved=True)
    .exclude(poster=self.user)
)
```
    .

carofun: Sorry Sam I didn't read and came myself exactly at the same time with a fixup I had trouble yesterday with this query and realised this morning I could avoid caring about anonymous users by requesting from users directly. Could you check my fixup to see if I already answered your comment

src/ashley/machina_extensions/forum_conversation/forms.py
```diff
@@ -0,0 +1,46 @@
...
+            .exclude(pk=self.user.pk)
+            .distinct()
+        )
+
+        list_active_users = [
+            {
+                "name": get_forum_member_display_name(user),
+                "link": reverse("forum_member:profile", kwargs={"pk": user.pk}),
+            }
+            for user in list(active_post_users)
```
sampaccoud, line 42:
```suggestion
            for user in active_post_users
```

src/ashley/machina_extensions/forum_conversation/forms.py
sampaccoud: :astonished: 
Ok so you are dropping the possibility to mention anonymous users. Maybe it will be confusing for users that some usernames can be mentioned and others not? I was thinking to allowing mentions and display them without any link. maybe it does not make sense :man_shrugging:?

carofun: At the moment we don't have this use case, in Django machina it's possible but not in Ashley. Each user are authenticated through LTI. It would be weird to deal with a case that we don't have and do extra work. It makes it more simple and realistic to ignore this use case. If we were allowing anonymous users to post messages in the future which is not certain, we still would be able to add them to mention them.

sampaccoud: True! :clap:

tests/ashley/machina_extensions/forum_conversation/test_forum_conversation.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # user2 become inactive
+        user2.is_active = False
+        user2.save()
+        # user3 loads the form for topic
+        form = PostForm(data=form_data, user=user3, forum=self.forum, topic=topic)
+        # We should only see user1 in the list of active users for this topic
+        assert form.fields["content"].widget.attrs["mention_users"] == [
+            {"name": "Benoit", "link": f"/forum/member/profile/{user1.id}/"},
+        ]
```
sampaccoud, line 198: This is really well written and robust and I'm fine merging it but maybe the reallybest practice would be to split it in x tests each proving one of the points. Each test would be simpler to understand and easier to maintain.

carofun: ok Thank you Sam for the review and feedback I will do so

tests/ashley/machina_extensions/forum_conversation/test_forum_conversation.py
```diff
@@ -37,162 +36,423 @@ def setUp(self):
...
+            },
+        ]
+
+    def test_list_active_users_only_concerns_users_with_approved_posts(self):
+        """
+        The form loads the list of active users for the current topic. We control that the list
+        only concerns users that have approved posts
+        """
+        user1 = UserFactory(
+            email="user1@mooc.fr",
```
madmatah, line 357: Why do you define an email address explicitly instead of letting the factory generate it?
(this comments apply for all your User instances in your tests).

In addition to that, I do not recommend to use a domain name that belong to someone else, it can lead to hazardous situations. I recommend to use domains reserved for test and documentation like `example.com` (see [IANA reserved domains](https://www.iana.org/domains/reserved) or [RFC2606](https://tools.ietf.org/html/rfc2606#section-2))

carofun: really good comment ! no point ! thank you

tests/ashley/machina_extensions/forum_conversation/test_forum_conversation.py
```diff
@@ -0,0 +1,447 @@
...
+    def test_list_active_users_only_concerns_active_users(self):
+        """
+        The form loads the list of active users for the current topic. We control that the list
+        only concerns users that have the status active
+        """
+        user1 = UserFactory(
+            lti_consumer=self.lti_consumer,
+            public_username="Benoit",
+            username="user1",
+        )
```
sampaccoud, line 395:
```suggestion
        user1 = UserFactory(
            lti_consumer=self.lti_consumer,
            public_username="Benoit",
            username="user1",
            is_active=False
        )
```

tests/ashley/machina_extensions/forum_conversation/test_forum_conversation.py
```diff
@@ -0,0 +1,447 @@
...
+        # user1 only see one time user1
+        assert form.fields["content"].widget.attrs["mention_users"] == [
+            {
+                "name": "Alfred",
+                "link": f"/forum/member/profile/{user2.id}/",
+                "username": "user2",
+            },
+        ]
+
+    def test_list_active_users_only_concerns_writter_of_current_topic(self):
```
sampaccoud, line 259:
```suggestion
    def test_list_active_users_only_concerns_writer_of_current_topic(self):
```

tests/ashley/machina_extensions/forum_conversation/test_forum_conversation.py
```diff
@@ -0,0 +1,447 @@
...
+        """
+        The form loads the list of active users for the current topic. We control that the list
+        only list writter of current topic and no other topic
```
sampaccoud, line 262:
```suggestion
        only contains writers involved in the current topic and no other users
```

src/ashley/machina_extensions/forum_conversation/forms.py
```diff
@@ -0,0 +1,38 @@
...
+            .order_by("poster__username")
+        )
+
+        # assign the list of active users for the current topic
+        self.fields["content"].widget.attrs["mention_users"] = [
+            {
+                "name": user["poster__username"],
+                "link": reverse(
+                    "forum_member:profile", kwargs={"pk": user["poster__id"]}
+                ),
```
madmatah: I'm wondering if this link should be generated here since it's not used in the AshleyEditor during the writing of the post by its author. The link is only used once the post has been rendered to HTML.

The link to the user profile is generated here and injected to the DraftJS component.
Then, if the post contains a mention, the link is stored in the draft.js JSON state that is POSTed to django-machina.
And then, it is stored in database without any validation.

We do not respect the #1 security rule here : **Don't trust user input**.

The Draft.js State can be modified by a user to inject a link to an external website (e.g. for SEO purpose) in a mention object, that will be rendered like a mention. The integrity between the user that we mention and the link that we render is not guaranteed.

An alternative solution to handle that can be to remove the link from the DraftJS state.
And generate the link it in `editor.decorator.mention`.

WDYT @carofun @sampaccoud ?
Do you think it's a problem we should address ?

carofun: _since it's not used in the AshleyEditor during the writing of the post by its author. The link is only used once the post has been rendered to HTML._

not to say you're not right, but juste to add a correction, actually the link is also used directly in the Editor,


 once you've added the @toto, you can see this step on the preview number 3 of this PR, we have the link already active here at this specific moment in the editor.

carofun: As a matter of fact as discussed, there is no point on being able to see this link in the Editor and as you mentioned we could use later on this information in the database thinking it's a reliable information, best thing to do is to build the link as you suggest on our own rendering the post.
I'll make this fix, thank you !

madmatah: > but the link is calculated in python how would you inject something malicious?

It is calculated in python to be injected in javascript into the the DraftJS editor.
During this step, you can do whatever you want with the draft.js state.

sampaccoud: True! Didn't catch that :fearful:

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
  - add frontend test for draft-js editor
  - add a ribbon icon when an instructor is the writer of a post or a topic
+ - add feature in draftjs editor to mention any active user in the current topic
```
madmatah, line 31:
```suggestion
 - add feature in AshleyEditor to mention any active user in the current topic
```
I suggest to use the name of our component instead of the technology it's based on.

And BTW, I realize that we should standardize the way we spell `Draft.js` in the Changelog (there is `Draft.js`, `draftjs`, `draft-js`). The official name of the project is `Draft.js`.

src/ashley/machina_extensions/forum_conversation/forms.py
carofun: Fixup has been released ! Thank you both

madmatah: Perfect! :1st_place_medal:

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -26,7 +36,7 @@ import {
   BlockquoteButton,
   CodeBlockButton,
 } from '@draft-js-plugins/buttons';
-
+import { MentionData } from '@draft-js-plugins/mention';
```
jbpenrath, line 39: You could remove this line and import `MentionData` above with `createMentionPlugin` and `defaultSuggestionFilter`

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -121,19 +132,55 @@ export const AshleyEditor = (props: MyEditorProps) => {
...
+
+    const { MentionSuggestions, mentionPlugin } = useMemo(() => {
+      // tslint:disable-next-line:no-shadowed-variable
+      const mentionPlugin = createMentionPlugin();
+      // tslint:disable-next-line:no-shadowed-variable
+      const { MentionSuggestions } = mentionPlugin;
+      return { mentionPlugin, MentionSuggestions };
+    }, []);
+
+    const onOpenChange = useCallback((_open: boolean) => {
```
jbpenrath, line 150: You remove this callback. As the boolean is the first parameter and the only argument you pass to `setOpen` you could use `setOpen` directly.

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -121,19 +132,55 @@ export const AshleyEditor = (props: MyEditorProps) => {
...
+      setOpen(_open);
+    }, []);
+    const onSearchChange = useCallback(({ value }: { value: string }) => {
+      setSuggestions(defaultSuggestionsFilter(value, props.mention_users!));
+    }, []);
+    plugins.push(mentionPlugin);
+    mentionPluginRender = (
+      <MentionSuggestions
+        open={open}
+        onOpenChange={onOpenChange}
```
jbpenrath, line 160:
```suggestion
        onOpenChange={setOpen}
```

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -121,19 +132,55 @@ export const AshleyEditor = (props: MyEditorProps) => {
...
+    const [open, setOpen] = useState(true);
+    const [suggestions, setSuggestions] = useState(props.mention_users);
+
+    const { MentionSuggestions, mentionPlugin } = useMemo(() => {
+      // tslint:disable-next-line:no-shadowed-variable
+      const mentionPlugin = createMentionPlugin();
+      // tslint:disable-next-line:no-shadowed-variable
+      const { MentionSuggestions } = mentionPlugin;
+      return { mentionPlugin, MentionSuggestions };
+    }, []);
```
jbpenrath, line 148: All this stuff to not write `<mentionPlugin.MentionsSuggestions />` ?

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -121,19 +132,55 @@ export const AshleyEditor = (props: MyEditorProps) => {
...
           editorState={editorState}
           onChange={editorChange}
-          plugins={[toolbarPlugin, emojiPlugin, linkPlugin, codeEditorPlugin]}
+          plugins={plugins}
           placeholder={props.placeholder}
           handleKeyCommand={keyBinding}
           handlePastedText={handlePastedText}
         />
         <emojiPlugin.EmojiSuggestions />
+        {mentionPluginRender}
```
jbpenrath, line 183: As you did for plugins, why not create a `PluginRenderers` array with default `[emojiPlugin.EmojiSuggestions]`, then you could push `mentionPluginRender` ?

And in your component do a 
```jsx
PluginRenderers.map(Plugin => <Plugin />)
```

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -35,6 +45,7 @@ interface MyEditorProps {
   target: HTMLInputElement;
   emojiConfig?: EmojiPluginConfig;
   linkPlaceholder?: string;
+  mention_users?: MentionData[];
```
jbpenrath, line 48: Snake case in da place ?! 🐍
```suggestion
  mentionUsers?: MentionData[];
```

carofun: true ! lint didn't see that one but JB did ! :-)

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -121,19 +131,44 @@ export const AshleyEditor = (props: MyEditorProps) => {
...
+    }, []);
+    plugins.push(mentionPlugin);
+    PluginRenderers.push(
+      <mentionPlugin.MentionSuggestions
+        key="mentions"
+        open={open}
+        onOpenChange={setOpen}
+        suggestions={suggestions}
+        onSearchChange={onSearchChange}
+        onAddMention={() => {}}
```
jbpenrath, line 154:
```suggestion
```

As this trigger do nothing, you could remove it.

jbpenrath (commented):
Looks good to me about frontend 👏

carofun (commented):
> Looks good to me about frontend 👏

kiool thanks for you feedback
I thought about it but WDYT of letting it in the code to know it exists. You find it useless ?... I'll get ride of it.
Thanks

src/ashley/editor/decorators.py
```diff
@@ -35,3 +35,27 @@ def emoji(props):
...
+
+    user = props.get("mention").get("user")
+    name = props.get("mention").get("name")
+
+    if name and user:
+        return DOM.create_element(
+            "a",
+            {
+                "class": "mention",
+                "href": f"/forum/member/profile/{user}/",
```
sampaccoud, line 56: I would find it more explicit to call it `user_id`:
```suggestion
                "href": f"/forum/member/profile/{user_id}/",
```

sampaccoud (approved):

src/ashley/editor/decorators.py
carofun: noted ! thanks

PR #164, opened 2021-03-08
🐛(frontend) handle error invalid configuration jest

carofun:
## Purpose

A runtime error happens loading lodash-es module and fails front tests. This
lib is published as untranspiled.

## Proposal
 We configure jest to allow transpiling this
specific module.


madmatah:
Resolved in #160

PR #171, opened 2021-04-02
promote and revoke users to moderator

carofun:
## Purpose

Students might want to give a hand in moderating forums of this LTIContext and teachers delighted to receive some support, as checking on all posts can take some time.
We need a feature to grant these privileges.

## Proposal

- Add new permission and group moderator 
- Add API to update users from frontEnd
- Add easy integration of new component in react and internationalisation
- Add interface in react to manage moderators

Capture d’écran 2021-04-02 à 18 08 15

Capture d’écran 2021-04-02 à 18 08 06

Capture d’écran 2021-04-02 à 18 13 57

Capture d’écran 2021-04-02 à 18 14 12

Capture d’écran 2021-04-02 à 18 14 07

Capture d’écran 2021-04-02 à 18 14 30

Capture d’écran 2021-04-02 à 18 14 36

Capture d’écran 2021-04-02 à 18 14 40


src/frontend/js/data/fetchApi.ts
```diff
@@ -0,0 +1,68 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+  } catch (error) {
+    return handle(error);
+  }
+  if (!response.ok) {
+    return handle(
+      new Error(`Failed to load list student : ${type} ${response.status}.`),
```
lunika, line 25:
```suggestion
      new Error(`Failed to load  ${type} list: ${response.status}.`),
```

src/frontend/js/data/fetchApi.ts
```diff
@@ -0,0 +1,68 @@
...
+    );
+  }
+  try {
+    const responseData = await response.json();
+    return responseData;
```
lunika, line 30:
```suggestion
    return await response.json();
```

src/frontend/js/data/fetchApi.ts
```diff
@@ -0,0 +1,68 @@
...
+  } catch (error) {
+    handle(new Error(`Failed to decode JSON in list ${type} ` + error));
```
lunika, line 32:
```suggestion
    handle(new Error(`Failed to decode JSON in list ${type}: ${error}`));
```

src/frontend/js/data/fetchApi.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+  } catch (error) {
+    return handle(error);
+  }
+  if (!response.ok) {
+    return handle(new Error(`Failed to update student : ${response.status}.`));
+  }
+  let responseData: User;
+  try {
+    responseData = await response.json();
```
lunika, line 61:
```suggestion
    return await response.json();
```

src/frontend/js/data/fetchApi.ts
```diff
@@ -0,0 +1,68 @@
...
+  } catch (error) {
+    return handle(
+      new Error('Failed to decode JSON in updateStudentToModerator ' + error),
+    );
+  }
+  return responseData;
```
lunika, line 67:
```suggestion
```

src/ashley/api/views.py
```diff
@@ -0,0 +1,114 @@
...
+User = get_user_model()
+LTIContext = get_model("ashley", "LTIContext")
+Forum = get_model("forum", "Forum")
+PermissionRequiredMixin: BasePermissionRequiredMixin = get_class(
+    "forum_permission.viewmixins", "PermissionRequiredMixin"
+)
+logger = logging.getLogger(__name__)
+
+
+class UserModeratorListApiView(APIView):
```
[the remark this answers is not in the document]
carofun: thank you I did a fixup to simplify the class.

src/ashley/api/serializers.py
```diff
@@ -0,0 +1,13 @@
...
+
+User = get_user_model()
+
+
+class UserSerializer(serializers.ModelSerializer):
+    """Serializer for User model."""
+
+    class Meta:
+        model = User
+        fields = ["id", "public_username"]
```
lunika, line 13: You should also add info about its role. 
```suggestion
        fields = ["id", "public_username", "role"]
        
    role = serializers.SerializerMethodField()
    
    def get_role(self, obj):
        """Return the role associated to this user (student, moderator, instructor)."""
        // your code here
```

src/ashley/api/views.py
```diff
@@ -0,0 +1,106 @@
...
+
+class UserModeratorListApiView(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """API requests to list moderators for current LTIContext"""
+
+    # Add permission to check if user has the right permission and can manage moderators
+    permission_classes = [ManageModeratorPermission]
+    serializer_class = UserSerializer
+
+    def get_queryset(self):
+        """List all the users that are moderators. Moderators have the primitive role
```
lunika, line 38:
```suggestion
        """Filters all the users that are moderators. Moderators have the primitive role
```

src/ashley/api/views.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def get_queryset(self):
+        "Gets all the students for the current LTIContext."
+        context = get_current_lti_session(self.request)
+        return User.objects.filter(
+            is_active=True,
+            groups__name=context.get_group_role_name(_FORUM_ROLE_STUDENT),
+        ).exclude(groups__name=context.get_group_role_name(_FORUM_ROLE_MODERATOR))
+
+    def update(self, request):
```
lunika, line 63: What is made in this method is wrong IMO.
It starts with the url patterns used and how you deal with it.
First of all the route should be a `PATCH` on  `/users/{user_id}` and in the body you should have an entire user object. So the role should also be in the serializer (I made a comment about that).

I also thinks that having 2 classes managing a user but with different roles is wrong.

classes `UserModeratorListApiView` and `UserStudentListApiView` share the same permission class, they return the same kind of list (a user). So I think you should only have one class `UserApiView`, in the list function you can filter the user you want by adding a querystring `role`. Without this querystring you return all the users related to this forum, if you want a specific role your should filter with the querystring like this `?role=[moderator,student]`, I wouldn't be surprise DRF can handle this case.

src/ashley/api/urls.py
```diff
@@ -0,0 +1,18 @@
...
+from .views import UserModeratorListApiView, UserStudentListApiView
+
+urlpatterns = [
+    path(
+        "students/",
+        UserStudentListApiView.as_view({"get": "list"}),
+        name="api_students",
+    ),
+    path(
+        "students/update/",
```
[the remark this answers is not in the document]
carofun: I avoided on purpose not to use standard REST update as I was finding the route not relevant. Using route /students/3/,  I would consider to update the student information. I was more thinking of coding an action, maybe I could rename update method into something else.
The aim is to add or remove group moderator to the student, I don't want to display details about the student, or to update the student information which I would expect to see on requesting students/{user_id}. Working around it again today, I can't figure out how to nicely deal with this.
WDYT if I rename this in students/update_moderator_role/ and leave it to POST.

src/ashley/api/urls.py
```diff
@@ -0,0 +1,18 @@
+"""Ashley API URLs"""
+from django.urls import path
+
+from .views import UserModeratorListApiView, UserStudentListApiView
+
+urlpatterns = [
+    path(
```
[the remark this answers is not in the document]
carofun: It still requires declaration, I've tried to use the generated url with a router but it defines more URL that I want to use so adjusting it like in the fixup targeting {"get": "list"} seems for this project to be a good way.

src/frontend/js/components/DashboardModerators/SelectStudentsSuggestField/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const propsFitered = props.users.filter((user) =>
+      user.public_username.toLowerCase().startsWith(text.trim().toLowerCase()),
+    );
+    return propsFitered.length > 0
+      ? propsFitered.slice(0, SUGGESTION_TO_SHOW)
+      : props.users.slice(0, SUGGESTION_TO_SHOW);
+  };
+
+  return (
+    <React.Fragment>
```
lunika, line 34: This fragment is useless, you cam remove it

src/frontend/js/data/frontEndData.ts
```diff
@@ -0,0 +1,12 @@
+import { handle } from '../utils/errors/handle';
+import { CommonDataProps } from '../types/commonDataProps';
+
+export const appFrontendContext: CommonDataProps['context'] = (window as any)
+  .__ashley_frontend_context__?.context;
+
+try {
+  if (!appFrontendContext)
+    throw new Error('No context frontend context available');
```
lunika, line 9: I don't see the interest to throw an error then catching it and then handle the error. Why not directly call `handle` ?
```suggestion
    handle(new Error('No context frontend context available'));
```

And then remove the `try... catch...` block

src/ashley/api/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+    role = serializers.SerializerMethodField()
+
+    class Meta:
+        model = User
+        fields = ["id", "public_username", "role"]
+        read_only_fields = ["public_username", "role"]
+
+    def get_role(self, obj):
+        """Describes the role of the user"""
+        context = get_current_lti_session(self.context.get("request"))
```
lunika, line 27: You should rename this variable because it is confusing with the serializer `context`

src/ashley/api/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+        if obj.is_active:
+            if obj.groups.filter(
+                name=context.get_group_role_name(_FORUM_ROLE_MODERATOR)
+            ).count() == 0 and obj.groups.filter(
+                name=context.get_group_role_name(_FORUM_ROLE_STUDENT)
+            ):
+                return _FORUM_ROLE_STUDENT
+            if obj.groups.filter(
```
lunika, line 35: It is not possible to do it in one queryset ? 

```suggestion
            if obj.groups.filter(name=context.get_group_role_name(_FORUM_ROLE_STUDENT), name=context.get_group_role_name(_FORUM_ROLE_STUDENT))
```

src/ashley/api/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+    class Meta:
+        model = User
+        fields = ["id", "public_username", "role"]
+        read_only_fields = ["public_username", "role"]
+
+    def get_role(self, obj):
+        """Describes the role of the user"""
+        context = get_current_lti_session(self.context.get("request"))
+        if obj.is_active:
+            if obj.groups.filter(
```
lunika, line 29: Should be great to do it in one queryset, maybe something like this ? (`Q` must be imported first `from django.db.models import Q`)

```suggestion
            if obj.groups.filter(name=context.get_group_role_name(_FORUM_ROLE_STUDENT), ~Q(name=context.get_group_role_name(_FORUM_ROLE_MODERATOR)))
```

src/ashley/api/views.py
```diff
@@ -0,0 +1,124 @@
...
+    # Add permission to check if user has the right permission and can manage moderators
+    permission_classes = [ManageModeratorPermission]
+    serializer_class = UserSerializer
+    ordering = "public_username"
+
+    def get_queryset(self):
+        context = get_current_lti_session(self.request)
+        role = self.request.GET.get("role", False)
+
+        if role is not False:
```
lunika, line 48: This `if` is useless and can be removed. Or you can do the opposite and then remove the block line 83:

```suggestion
        if role is False:
            # If no role is specified, filter all the users for the current LTIContext
            return (
                User.objects.filter(
                    groups__name__startswith=context.get_group_role_name("")
                )
                .order_by(self.ordering)
                .distinct()
            )
```        
```

File: src/ashley/api/views.py [hunk 18]
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+    # pylint: disable=W0221
+    def update(self, request, pk=None):
+        """Update method is used to promote or revoke moderator role on student users"""
+        user = User.objects.get(pk=pk)
+        context = get_current_lti_session(self.request)
+        group_moderator_name = context.get_group_role_name(_FORUM_ROLE_MODERATOR)
+        group_student_name = context.get_group_role_name(_FORUM_ROLE_STUDENT)
+        try:
```
lunika, line 98: Can you add a comment explaining what you are doing in this `try` block ?

src/ashley/api/views.py
```diff
@@ -0,0 +1,124 @@
...
+            group_moderator = Group.objects.get(name=group_moderator_name)
+        except Group.DoesNotExist as no_group_moderator:
+            raise PermissionDenied() from no_group_moderator
+
+        # Role asked is student, check user is both student and moderator before removing the group
+        if (
+            self.request.data.get("role") == _FORUM_ROLE_STUDENT
+            and user.groups.filter(name=group_moderator_name)
```
lunika, line 106: If it works in the serializer, you should use only one queryset.

src/ashley/api/views.py
```diff
@@ -0,0 +1,124 @@
...
+            and user.groups.filter(name=group_student_name)
+        ):
+            user.groups.remove(group_moderator)
+            user.save()
+        # Role asked is moderator, check user is student and not yet moderator before adding
+        # the group moderator
+        elif (
+            self.request.data.get("role") == _FORUM_ROLE_MODERATOR
+            and user.groups.filter(name=group_student_name)
```
lunika, line 115: Same here

src/ashley/api/views.py
```diff
@@ -0,0 +1,124 @@
...
+            and user.groups.filter(name=group_moderator_name).count() == 0
+        ):
+            user.groups.add(group_moderator)
+            user.save()
+        else:
+            logger.debug("Query is not expected")
```
lunika, line 121: Which query ? I think this message is not explicit, alone in a log file I will never understand what is the purpose of this message without reading the code.

src/ashley/context_mixins.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def get_current_lti_session_first_forum(request):
+    """
+    Gets an eligible forum from current LTIContext object.
+    A LTIContext can have multiple forums.
+    """
+    if request.user.is_authenticated and request.session.get(SESSION_LTI_CONTEXT_ID):
+        context = get_current_lti_session(request)
+        forum = Forum.objects.filter(lti_contexts=context).first()
```
lunika, line 27:
```suggestion
        return Forum.objects.filter(lti_contexts=context).first()
```

src/ashley/context_mixins.py
```diff
@@ -0,0 +1,30 @@
...
+        return forum
```
lunika, line 28:
```suggestion
```

lunika (commented):
Do we need to have a user in both group `student` and `moderator` ? I don't know what it implies to just have a user in a single group but I don;t really see the interest to have a user in a `student` group **and** in a `moderator` group.

We could simply have this hierarchy group ?

- user
- moderator
- instructor

WDYT ?

carofun:
> Do we need to have a user in both group `student` and `moderator` ? I don't know what it implies to just have a user in a single group but I don;t really see the interest to have a user in a `student` group **and** in a `moderator` group.
> 
> We could simply have this hierarchy group ?
> 
> * user
> * moderator
> * instructor
> 
> WDYT ?

Moderator group is an internal group, student group and instructor groups are used from LTI Roles. There is anonymous users has well in the forum. Our moderators are necessary users that have the student role in LTI, this is a pre-required.
For this project we have not created a new LTI role because we can't, we have just decorated student with a new group permission Moderator if they are supposed to be moderators.

src/frontend/js/data/fetchApi.ts
carofun: I haven't done this one because if I don't return User type, it upsets code.

src/ashley/api/views.py
carofun: lint made me do it this way.
It wants a return statement and it bothers me to do a return statement at the end for a default case which means role is False letting the code testing 3 times role for nothing.

src/frontend/js/data/fetchApi.ts
```diff
@@ -0,0 +1,70 @@
...
+    response = await fetch(`/api/v1/users/${user.id}/`, {
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRFTOKEN': appFrontendContext.csrftoken,
+      },
+      body: JSON.stringify(user),
+      method: 'PUT',
+    });
+  } catch (error) {
+    return handle(error);
```
lunika, line 58: If it fails, this function silently fail. You are handling the error but the caller never know the api call is failing.

src/frontend/package.json
```diff
@@ -20,53 +22,73 @@
   "author": "France Université Numérique",
   "license": "MIT",
   "devDependencies": {
-    "@babel/core": "7.13.14",
+    "@babel/core": "7.13.10",
```
lunika, line 25: Most of the change you made are downgrading the installed version. Is it a copy pasta fails ?

src/frontend/tsconfig.json
```diff
@@ -17,12 +17,15 @@
...
+    "noImplicitAny": true,
+    "noImplicitThis": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
+    "strictNullChecks": true,
     "sourceMap": true,
     "target": "es6",
-    "types": ["jest"]
+    "types": ["node"]
```
lunika, line 28: Why ?

src/ashley/api/views.py
carofun: so quiet

src/ashley/api/serializers.py
carofun: didn't manage to do it with Q or with combined filters, I think it's because I'm checking on many to many values, anyway i changed the code to only get one query by getting once the group list and just checking if group is part of it.

src/ashley/api/views.py
lunika: And something like this ? 

```
def get_queryset(self):
        context = get_current_lti_session(self.request)
        role = self.request.GET.get("role", False)

        # Filter all the users that are moderators or students."""
        if role == _FORUM_ROLE_STUDENT:
            # Gets all the students for the current LTIContext.
            return (
                User.objects.filter(
                    is_active=True,
                    groups__name=context.get_group_role_name(_FORUM_ROLE_STUDENT),
                )
                .exclude(
                    groups__name=context.get_group_role_name(_FORUM_ROLE_MODERATOR)
                )
                .order_by(self.ordering)
            )
        if role == _FORUM_ROLE_MODERATOR:
            # Gets all the moderators for the current LTIContext.Moderators have
            # the primitive role student."""
            return (
                User.objects.filter(
                    is_active=True,
                    groups__name=context.get_group_role_name(_FORUM_ROLE_MODERATOR),
                )
                .filter(groups__name=context.get_group_role_name(_FORUM_ROLE_STUDENT))
                .order_by(self.ordering)
            )
        if role == _FORUM_ROLE_INSTRUCTOR:
            # Gets all the instructor for the current LTIContext.
            return User.objects.filter(
                is_active=True,
                groups__name=context.get_group_role_name(_FORUM_ROLE_INSTRUCTOR),
            ).order_by(self.ordering)

        return (
            User.objects.filter(
                groups__name__startswith=context.get_group_role_name("")
            )
            .order_by(self.ordering)
            .distinct()
        )
```

src/frontend/package.json
carofun: might be a result of a no conflict

src/frontend/js/data/fetchApi.ts
carofun: Do you have a suggestion? I didn't know what to do with it

lunika: I suggest you to just throw the error. And it's the caller job to handle it. Maybe it wants to `report` the error ? Do an alternative call, etc.

src/ashley/api/views.py
carofun: it's what I was telling you, our code will test three times the value of role when it's not even defined it it is the case, that was the point that was bothering me

lunika: I think it's clearer to limit the number of `if` indentation. the `if role is not False` is really useless I think.

src/frontend/js/components/DashboardModerators/ListModerators/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+              values={{
+                moderatorCount: props.users.length,
+                userCount: props.totalUsers,
+              }}
+            />
+          </div>
+          {props.users.map((user, key) => {
+            return (
+              <Moderator
+                key={`${user.id}${key}`}
```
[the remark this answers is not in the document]
carofun: it should

src/frontend/js/components/DashboardModerators/Modal/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </div>
+        <div className="modal-body">
+          <FormattedMessage
+            {...messagesDashboardModerators.modalConfirmation}
+          />
+        </div>
+        <div className="modal-footer">
+          {props.user && (
+            <ButtonChangeRoleCta
+              key="{props.user.id}+{props.action}"
```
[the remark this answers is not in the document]
carofun: none thank you, i had misunderstood the key needs.

src/frontend/js/components/DashboardModerators/ListModerators/index.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+  setUser: (user: User) => {
+    console.warn(
+      'setUser has been called for user ' +
+        user.id +
+        ' ' +
+        user.public_username,
+    );
+  },
+  onChange: () => {
+    console.warn('OnChange has been called for sure !');
```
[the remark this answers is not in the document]
carofun: I didn't know I thought it was the same than french, thank you for the tip !

src/frontend/js/components/DashboardModerators/common.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export interface ListUserProps {
+  users: User[];
+  onChange: () => void;
+  setUser: (value: User) => void;
+}
+export interface ListModeratorProps extends ListUserProps {
+  totalUsers: number;
+}
+export interface ModalDashboardModeratorProps {
```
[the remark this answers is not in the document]
carofun: and if I have another modal in the project later on ? It's probably going to happen, I wanted a type specfic to this modal. WDYT ?

src/frontend/js/components/DashboardModerators/index.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+    { public_username: 'Zoé', id: 8 },
+  ]);
+  fetchMock.mock('/api/v1/users/6/', {});
+  // create the element for the setAppElement used in the Modal
+  const appContainer = document.createElement('div');
+  appContainer.setAttribute('id', 'modal-exclude__react');
+  document.body.append(appContainer);
+
+  it('renders the SelectStudentsSuggestField component called in this component', async () => {
+    await act(async () => {
```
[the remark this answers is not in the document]
carofun: Bingo, You are absolutely right about it !!

src/frontend/js/components/DashboardModerators/index.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+      name: /revoke moderator/i,
+    });
+    // Modal component is loaded and operational from ListModerator
+    fireEvent.click(button);
+    await screen.findByRole('heading', { name: 'Revoke Samy to moderator' });
+    await screen.findByText(
+      'Are you certain you want to confirm this action ?',
+    );
+    const closed = await screen.findAllByRole('button', { name: 'Close' });
+    fireEvent.click(closed[0]);
```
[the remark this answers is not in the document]
carofun: true !It would be nice to control it's really closed.

src/frontend/js/components/DashboardModerators/common.ts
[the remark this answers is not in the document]
lunika: If the interface is defined in the component file, then you will have no conflict. If you leave it here, you will have this kind of conflict.

src/frontend/package.json
carofun: I'll have to be careful in my next rebase

src/frontend/js/components/DashboardModerators/common.ts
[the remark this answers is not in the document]
carofun: got it thank you

src/frontend/tsconfig.json
carofun: I use require in the tests and if I don't change jest config, it returns me this error, 7 times as I use require 7 times
Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add `node` to the types field in your tsconfig.
require('.').default

src/frontend/js/components/DashboardModerators/ButtonChangeRoleCta/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+      : messagesDashboardModerators.revokeModerator;
+
+  const updateStudent = async () => {
+    // Depending on action submit the requested role
+    switch (action) {
+      case Actions.PROMOTE:
+        user.role = Role.MODERATOR;
+        break;
+      case Actions.REVOKE:
+        user.role = Role.STUDENT;
```
lunika, line 31: I know here it's not needed but adding a `break` for each `case` statement is I think a good practice to avoid future mistake

src/frontend/js/components/DashboardModerators/ListModerators/index.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+  },
+  onChange: () => {
+    console.warn('OnChange has been called for sure!');
+  },
+};
+
+describe('<ListModerators />', () => {
+  const contextProps = ContextFactory().generate();
+  (window as any).__ashley_frontend_context__ = { context: contextProps };
+  const ListModerators = require('.').default;
```
lunika, line 28: Why are you using `require` here ? Why not adding `import ListModerators from '.';` with the other `import` ?

src/frontend/js/components/DashboardModerators/ButtonChangeRoleCta/index.spec.tsx
```diff
@@ -0,0 +1,73 @@
+import 'regenerator-runtime/runtime';
```
lunika, line 1:
```suggestion
```

this should not be done here. You just fix the use of regenerator on this test. But in fact regenerator must be used globally and not only for tests because babel will need it to transpile to the targets you defined.
2 things you have to do:

- add `'regenerator-runtime/runtime'` in your webpack entry section
- add `'regenerator-runtime/runtime'` in your jest `setupFilesAfterEnv` section

Also, most of time the same thing is made when you use core-js

src/frontend/js/components/DashboardModerators/ButtonChangeRoleCta/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { messagesDashboardModerators } from '../messages';
+import { Actions, Role } from '../../../types/Enums';
+import { User } from '../../../types/User';
+import { updateUser } from '../../../data/fetchApi';
+
+export interface ButtonModeratorCtaProps {
```
lunika, line 8:
```suggestion
export interface ButtonChangeRoleCtaProps {
```

src/frontend/js/components/DashboardModerators/ListModerators/index.spec.tsx
```diff
@@ -0,0 +1,85 @@
+import 'regenerator-runtime/runtime';
```
lunika, line 1:
```suggestion
```

src/frontend/js/components/DashboardModerators/ListModerators/index.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { User } from '../../../types/User';
+import { Moderator } from '../Moderator';
+import { messagesDashboardModerators } from '../messages';
+import { ListUserProps } from '../SelectStudentsSuggestField';
+
+export interface ListModeratorProps extends ListUserProps {
```
lunika, line 8:
```suggestion
export interface ListModeratorsProps extends ListUserProps {
```

And I don't think it's a good practice to extends component props from an other component props. Specially in that case. this component is not a child of `SelectStudentsSuggestField` component.

src/frontend/js/components/DashboardModerators/Modal/index.spec.tsx
```diff
@@ -0,0 +1,121 @@
+import 'regenerator-runtime/runtime';
```
lunika, line 1:
```suggestion
```

src/frontend/js/components/DashboardModerators/Moderator/index.spec.tsx
```diff
@@ -0,0 +1,45 @@
+import 'regenerator-runtime/runtime';
```
lunika, line 1:
```suggestion
```

src/frontend/js/components/DashboardModerators/SelectStudentsSuggestField/index.spec.tsx
```diff
@@ -0,0 +1,114 @@
+import 'regenerator-runtime/runtime';
```
lunika, line 1:
```suggestion
```

src/frontend/js/components/DashboardModerators/SelectStudentsSuggestField/index.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState } from 'react';
+import { FormattedMessage, useIntl } from 'react-intl';
+import AutoSuggest from 'react-autosuggest';
+import { messagesDashboardModerators } from '../messages';
+import { User } from '../../../types/User';
+
+const SUGGESTION_TO_SHOW = 5;
+
+export interface ListUserProps {
```
lunika, line 9:
```suggestion
export interface SelectStudentsSuggestFieldProps {
```

src/frontend/js/components/DashboardModerators/index.spec.tsx
```diff
@@ -0,0 +1,117 @@
+import 'regenerator-runtime/runtime';
```
lunika, line 1:
```suggestion
```

src/frontend/js/components/DashboardModerators/ButtonChangeRoleCta/index.spec.tsx
carofun: I included it in index.tsx, webpack is indeed a much nicer way, thank you for the feedback

src/frontend/js/components/DashboardModerators/ButtonChangeRoleCta/index.tsx
carofun: now that props are in the same file :-) it would be nicer !

src/frontend/js/components/DashboardModerators/ListModerators/index.tsx
carofun: When it was in the same file, I thought it was a good idea in order not to repeat yourself, now I don't see the point anymore for comprehension.

src/frontend/js/components/DashboardModerators/SelectStudentsSuggestField/index.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+    console.warn(
+      `setUser has been called for user ${user.id} ${user.public_username}`,
+    );
+  },
+  onChange: jest.fn(),
+};
+
+describe('<SelectStudentsSuggestField />', () => {
+  beforeEach(() => {
+    render(
```
lunika, line 29: Same here, no render in a beforeEach.

tests/ashley/api/test_api_manage_moderators.py
```diff
@@ -0,0 +1,843 @@
...
+                f"{lti_context.base_group_name}:role:student",
+                f"{lti_context.base_group_name}:role:moderator",
+            ],
+            list(update_user.groups.values_list("name", flat=True)),
+        )
+
+    def test_api_can_manage_moderators_update_student_id_param_ignored(
+        self,
+    ):
+        """If id in body of request api is different than the id in the url is ignored.
```
sampaccoud, line 791:
```suggestion
        """If id in body of request api is different from the id in the url, it is ignored.
```

src/frontend/js/components/DashboardModerators/index.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+      'Are you certain you want to confirm this action ?',
+    );
+
+    const promoteModerator = await screen.findByRole('button', {
+      name: 'Promote moderator',
+    });
+    await waitFor(() => {
+      fireEvent.click(promoteModerator);
+    });
+    expect(fetchMock.called('/api/v1/users/6/')).toEqual(true);
```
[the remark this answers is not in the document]
carofun: i left one by mistake

src/ashley/urls.py
```diff
@@ -15,5 +16,6 @@
         name="forum.username.change",
     ),
     path("moderators/", ManageModeratorsView.as_view(), name="moderators"),
+    path("api/v1/", include(api_urls)),
```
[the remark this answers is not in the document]
carofun: I dit half of the job, for unicity i kept the format so now url are changed to v1.0 as Richie but using the regular exprssion bloked ashley on the middleware for the update method. I've tried to fix it with no success, so i've left v1.0 and got rid of the regular expression 

Capture d’écran 2021-04-16 à 15 57 15

lunika (approved):
LGTM

Few comments:

- `console.warn` is not the appropriated way to test if a function is called, `jest.mock` is the right way to do that.
- Failing scenarios are missing, you should add them when working on #176

madmatah:
Great feature!

I just tested it and found a small bug. On the moderator list of a forum, the link that points to user profiles is invalid:

invalid link

.gitignore
```diff
@@ -27,6 +27,8 @@ share/python-wheels/
 *.egg
 MANIFEST
 
+# Exported strings & generated translation templates
+src/frontend/i18n/frontend.json
```
madmatah, line 31: When I execute `make build-front`, I end up with non-versioned files : 
```
$ git status
On branch feat/moderator-cleaning
Your branch is up to date with 'origin/feat/moderator-cleaning'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   src/frontend/js/translations/es-ES.json
        modified:   src/frontend/js/translations/fr-CA.json
        modified:   src/frontend/js/translations/fr-FR.json
        modified:   src/frontend/js/translations/fr.json
```

Maybe something is missing in the `.gitignore` ?

src/frontend/i18n/locales/es-ES.json
```diff
@@ -0,0 +1,38 @@
+{
```
madmatah, line 1: How did you chose the locales list ?

At this time, according to our [Crowdin project ](https://crowdin.com/project/ashleyforum), the source language is English and the project is translated to French (fr_FR) only. I'll be happy to see Ashley translated in other languages, but I think we should be consistent between the backend and the frontend translations.

.gitignore
carofun: with the make lint-front it gets sorted

src/frontend/i18n/locales/es-ES.json
carofun: I only left french file

carofun:
> Great feature!
> 
> I just tested it and found a small bug. On the moderator list of a forum, the link that points to user profiles is invalid:
> 
> invalid link

Good catch, I totally forgotten I initially put such a link on this item. I've added translation and tests and fixed the link.

.gitignore
carofun: Files are now ignored it git !

src/ashley/api/views.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # If no role is specified, filter all the users for the current LTIContext
+        return (
+            User.objects.filter(
+                is_active=True,
+                groups__name=context.base_group_name,
+            )
+            .order_by(self.ordering)
+            .distinct()
+        )
```
madmatah, line 99: I think this function can be factored to avoid repeating treatments on each case (e.g. filtering active_users and sorting).

Also, I think that you are trying to handle an exhaustive list of cases that cannot predicted due to the fact that LTI role names are not standardized and depends of the LTI Consumer implementation.

I suggest the following rewrite of the function, with these notable changes:
 - I removed the if `if role == _FORUM_ROLE_INSTRUCTOR` and `if role == _FORUM_ROLE_STUDENT`
 - I generalized the use of the `!` prefix to exclude a role from the results to every possible roles

```suggestion
    def get_queryset(self):
        context = get_current_lti_session(self.request)
        role = self.request.GET.get("role", None)

        # Select all active users who are part of the current LTIContext
        query = User.objects.filter(
            is_active=True,
            groups__name=context.base_group_name,
        )

        # If a role is specified and start with !, we exclude its members from the results
        if role is not None and role.startswith("!"):
            query = query.exclude(groups__name=context.get_group_role_name(role[1:]))
        # If a role is specified, we select only users having this role inside the LTIContext
        elif role:
            query = query.filter(groups__name=context.get_group_role_name(role))

        # When the role moderator (or !moderator) is specified, we want to exclude the
        # users having the instructor role from the results.
        if role in [_FORUM_ROLE_MODERATOR, f"!{_FORUM_ROLE_MODERATOR}"]:
            query = query.exclude(groups__name=context.get_group_role_name(_FORUM_ROLE_INSTRUCTOR))

        return query.order_by(self.ordering)
```

src/ashley/api/serializers.py
```diff
@@ -0,0 +1,27 @@
...
+        model = User
+        fields = ["id", "public_username", "role"]
+        read_only_fields = ["id", "public_username", "role"]
+
+    def get_role(self, obj):
+        """Describes the role of the user"""
+        if obj.is_active:
+
+            lti_context = get_current_lti_session(self.context.get("request"))
+            return lti_context.get_user_roles(obj)
```
madmatah, line 25: I also suggest to use the plural form here:

```suggestion
    roles = serializers.SerializerMethodField()

    class Meta:
        model = User
        fields = ["id", "public_username", "roles"]
        read_only_fields = ["id", "public_username", "roles"]

    def get_roles(self, obj):
        """Returns the role list of the user within the current LTIContext."""
        if obj.is_active:

            lti_context = get_current_lti_session(self.context.get("request"))
            return lti_context.get_user_roles(obj)
```

src/ashley/defaults.py
```diff
@@ -9,6 +9,8 @@
 from ashley.editor.decorators import emoji, link, mention
 
 _FORUM_ROLE_INSTRUCTOR = "instructor"
+_FORUM_ROLE_STUDENT = "student"
```
madmatah, line 12:
```suggestion
```

I suggest to remove this constant.

src/ashley/api/views.py
carofun: much better !

madmatah (approved):

PR #175, opened 2021-04-13
✨(forum) sort discussion topics

### LINKED ISSUE #66: Allow sorting discussion topics
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
On the view listing discussion topics, we want to allow sorting topics by each column : number of answers, number of views and date.

topics

**Describe the solution you'd like**
I propose to follow the way it is done in Django's admin using a querystring parameter.

Make the column headers clickable. When a column header is clicked, the value in this column is added as first sorting criteria and 2 buttons are added to the column header: remove from sorting and toggle sorting.

The sorting critera are set in the querystring (Django admin uses the "o" parameter) as a dot-separated list of the ids of columns that are active for sorting, in order of priority (last clicked first). Each column id is preceded by a minus sign if it should be sorted by descending order. 

For example a querystring parameter `o=6.-7` will sort first by column id 6 in ascending order, then by column 7 in descending order.

A click on "remove sorting" removes this column from sorting priorities
A click on "toggle sorting" switched between ascending and descending order.

**Describe alternatives you've considered**
In fact, this should probably be proposed as a PR to https://github.com/ellmetha/django-machina

**Discovery, Documentation, Adoption, Migration Strategy**
While preparing your pull request, follow guidelines in our [handbook](https://openfun.gitbooks.io/handbook/content/) and look at examples of pull requests that have already been merged to see how we do it.

carofun:
## Purpose

The view listing the different topics in the forum is by default ordered by last messages which enables to see recent activity. If we want to see which subjects in the past were dynamic, it’s interesting to be able to sort discussions from other parameters such as number of replies or views. 

## Proposal

Make the column headers clickable. When a column header is clicked, the column is sorting by this column. Once it's sorted, we can change the sens of the sorting. By default it's sorting on last post date.

By default, there is no parameter in the url and the first topic shown is the one having the most recent activity :
Capture d’écran 2021-04-13 à 11 05 17

We've clicked on the arrow to reverse the order
Capture d’écran 2021-04-13 à 11 05 26

We click on a new colum
Capture d’écran 2021-04-13 à 11 05 38

And reverse the order
Capture d’écran 2021-04-13 à 11 05 46

We click on a new colum
Capture d’écran 2021-04-13 à 11 06 01

Resolve #66


src/ashley/machina_extensions/forum/views.py
```diff
@@ -15,20 +17,132 @@
...
+        """
+        param = self.params.copy()
+        for key, value in new_params.items():
+            if value is None:
+                if key in param:
+                    del param[key]
+            else:
+                param[key] = value
+
+        return "?%s" % urlencode(sorted(param.items()))
```
madmatah, line 126: Unless I am mistaken, I don't see any case in which these lines are executed : 
```python
            if value is None:
                if key in param:
                    del param[key]
```
If my assumption is correct, I suggest the following rewriting of this function:

```suggestion
        param = {**self.params, **new_params}
        return "?%s" % urlencode(sorted(param.items()))
```

src/ashley/machina_extensions/forum/views.py
```diff
@@ -15,20 +17,132 @@
...
+        return query.order_by(self.get_ordering())
+
+    def get_context_data(self, **kwargs):
+        """Returns the context data to provide to the template."""
+        context = super().get_context_data(**kwargs)
+        # add sort header links containing asc, desc orders
+        context["header"] = list(self.result_headers())
+        # used for pagination to keep the order
+        if self.request.GET.get(ORDER_VAR):
+            context["order"] = f"{ORDER_VAR}={self.request.GET.get(ORDER_VAR)}"
```
madmatah, line 84: The value injected in the context without validation.
If an invalid value is provided, it will still be used to build pagination links, as you can see in the following example:

Screenshot from 2021-04-21 15-40-59

carofun: True, I didn't think it could be a problem, but you right why allowing it! thanks

src/ashley/machina_extensions/forum/views.py
carofun: I wish I had done this one, nice 👍

src/ashley/machina_extensions/forum/views.py
```diff
@@ -15,20 +17,132 @@
...
 )
 
 
+class ForumView(BaseForumView):  # pylint: disable=too-many-ancestors
+    """Displays a forum and its topics."""
+
+    # header columns
+    list_display = ["subject", "posts_count", "views_count", "last_post_on"]
+    # default ordering
+    ordering = ["-last_post_on"]
```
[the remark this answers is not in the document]
carofun: I only did it because in django's view I read it was the convention. Am I wrong?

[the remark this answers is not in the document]
madmatah: Wasn't it with for views with multi sort criteria?
I don't know any convention about this, it was just my opinion :)

madmatah (approved):

PR #178, opened 2021-04-19
🚚(frontend) clean built frontend files before each build

carofun:
## Purpose

Webpack uses hashes to name files it generates. But currently, old files stay in place. So after several builds, we can have a js folder with a very large amount of files which is not efficient and furthermore can produce issues and slow down script executions.

In order to clean these files before each build, we need to put them in a dedicated directory to not remove
other js scripts placed in `static/js` directory.

To optimize the build of docker, we ignore `assets` files.

## Proposal

- [x] Generate frontend scripts into a dedicated `build` folder
- [x] Clean this new folder before each build
- [x] Ignore assets files for docker

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -359,7 +359,7 @@ jobs:
           name: List builds
           command: |
             echo "== Javascript =="
-            ls ../ashley/static/ashley/js/*.js
+            ls ../ashley/static/ashley/js/build/*.js
             echo "== CSS =="
             ls ../ashley/static/ashley/css/*.css
             echo "== Webfonts =="
@@ -367,7 +367,7 @@ jobs:
       - persist_to_workspace:
           root: ~/fun
           paths:
-            - src/ashley/static/ashley/js/*.js
+            - src/ashley/static/ashley/js/**/*.js
             - src/ashley/static/ashley/css/*.css
             - src/ashley/static/ashley/font/*
   lint-front:
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,3 +29,8 @@ db.sqlite3
 
 # Font-end
 **/node_modules
+
+# Assets
+src/ashley/static/ashley/js
+src/ashley/static/ashley/css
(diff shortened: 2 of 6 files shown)


.dockerignore
```diff
@@ -29,3 +29,7 @@ db.sqlite3
 
 # Font-end
 **/node_modules
+
+# Assets
+src/ashley/static/ashley/js
+src/ashley/static/ashley/css
```
madmatah, line 35: Maybe you want to add the `fonts` directory, since it's also generated by the frontend ?

carofun: true, thank you

.circleci/config.yml
```diff
@@ -359,15 +359,15 @@ jobs:
...
+            ls ../ashley/static/ashley/js/build/*.js
             echo "== CSS =="
             ls ../ashley/static/ashley/css/*.css
             echo "== Webfonts =="
             ls ../ashley/static/ashley/font/
       - persist_to_workspace:
           root: ~/fun
           paths:
-            - src/ashley/static/ashley/js/*.js
+            - src/ashley/static/ashley/js/build/*.js
```
jbpenrath, line 370: You could merge these two lines with:
```
- src/ashley/static/ashley/js/**/*.js
```

Dockerfile
```diff
@@ -44,8 +44,8 @@ COPY ./src/ashley /builder/src/ashley/
 
 # Copy distributed application's statics
 COPY --from=front-builder \
-    /builder/src/ashley/static/ashley/js/*.js \
-    /builder/src/ashley/static/ashley/js/
+    /builder/src/ashley/static/ashley/js/build/*.js \
+    /builder/src/ashley/static/ashley/js/build/
```
jbpenrath, line 48: You could copy the whole folder nope?

```
/builder/src/ashley/static/ashley/js \
/builder/src/ashley/static/ashley/js
```

jbpenrath (approved):
LGTM

PR #180, opened 2021-04-21
🚨(linter) exclude Ashley's static directory from mypy config

carofun:
## Purpose

From time to time, mypy crashes with the error Input/output error src/ashley/static/ashley/js. As mypy use is irrelevant in this target directory, we choose to exclude this directory from mypy configuration.

## Proposal

Exclude src/ashley/static/ashley/ from mypy configuration


setup.cfg
```diff
@@ -100,6 +100,7 @@ testpaths =
 
 [mypy]
 ignore_missing_imports = True
+exclude = src/ashley/static/ashley/
```
madmatah, line 103:
```suggestion
exclude = src/ashley/static/
```

madmatah (approved):

PR #188, opened 2021-05-26
Feat/draft upload image

carofun:
## Purpose

Decorating posts with images will help students and teachers communicate.
Adding images it's an expected feature in a forum, and it's user-friendly. 
Images can take a lot of space and resources. We want to use an external
service for this job.

## Proposal

Adding upload image feature on forum

- Adding internationalization on AshleyEditor not to send text through props
- Configuring Amazon S3 for the project
- Add new endpoint on DFR to upload images and save them on Amazon, associate image with forum and user in a new model for cleaning issues
- Add a modal in AshleyEditor to be able to upload images
- Add draft-js plugin to play around with images such as alignment, resize, and move the image by drag'n drop in the editor and change css property so that it's the most alike between the preview in the editor and the final result of the post.

https://user-images.githubusercontent.com/76939263/119986841-a5239580-bfc4-11eb-808a-6e176ed7bf69.mov

https://user-images.githubusercontent.com/76939263/119956707-aabbb400-bfa1-11eb-8bb5-7f7602f88b8d.mov


src/ashley/models.py
```diff
@@ -197,3 +212,50 @@ class Meta:
...
+        max_length=100,
+        blank=True,
+        validators=[
+            validate_upload_image_file_size,
+            validate_upload_image_allowed_type,
+        ],
+        null=True,
+    )
+    forum = models.ForeignKey(
+        "forum.Forum",
```
madmatah, line 238:
```suggestion
        get_model("forum", "Forum"),
```

With this change (+ the correct import of `get_model`), you can remove your `pylint: disable-all` comment at the beginning of this module.

The error you had (`No module named forum.models`) is related to the forum Foreign key.
The forein key checker of pylint-django doesn't like strings, it seems to prefer class references.

src/ashley/models.py
```diff
@@ -197,3 +212,50 @@ class Meta:
...
+        on_delete=models.CASCADE,
+        verbose_name=_("Forum"),
+        blank=True,
+        null=True,
+    )
+    poster = models.ForeignKey(
+        get_user_model(),
+        blank=True,
+        null=True,
```
madmatah, line 247: Why is this field optional?
For which use case?

src/ashley/models.py
```diff
@@ -197,3 +212,50 @@ class Meta:
...
+        on_delete=models.CASCADE,
+        verbose_name=_("Poster"),
+    )
+
+    class Meta:
+        """Options for the `AbstractImage` model."""
+
+        abstract = True
+        app_label = "ashley"
+        ordering = ["-created", "id"]
```
madmatah, line 257: Can you explain in a code comment why this is necessary?

src/ashley/permissions.py
```diff
@@ -55,3 +55,17 @@ def perform_permissions_check(self, user, obj, perms):
...
+
+    # pylint: disable=R0201,unused-argument
+    def has_permission(self, request, view):
+        """Check user is authenticated"""
+        return request.user.is_authenticated
+
+    # pylint: disable=R0201,unused-argument
+    def has_object_permission(self, request, view, obj):
+        """Check user is authenticated"""
+        return request.user.is_authenticated
```
madmatah, line 71: Django Rest Framework provides a permission class called `IsAuthenticated`.
Unless I missed something, you should use it instead of this class that does the same.

src/ashley/validators.py
```diff
@@ -39,3 +41,20 @@ def error(self, message):
     def handle_data(self, data):
         """Concatenate the plain text found in each element."""
         self.text += data
+
+
+def validate_upload_image_file_size(file):
+    """Controls the size of the uploaded file is not over the limit authorized"""
+    if file.size > settings.MAX_UPLOAD_FILE_MB * 1000 * 1024:
```
madmatah, line 48:
```suggestion
    if file.size > settings.MAX_UPLOAD_FILE_MB * 1024 * 1024:
```

madmatah (commented):
Thanks, we were looking forward to this feature with great interest! :star_struck: 
Here is a first serie of feedbacks on the backend.

src/frontend/js/components/AshleyEditor/index.tsx
```diff
@@ -104,7 +143,9 @@ export const AshleyEditor = (props: MyEditorProps) => {
   };
 
   const editorChange = (stateEditor: EditorState) => {
-    props.target.value = JSON.stringify(
+    (document.getElementById(
+      props.target,
+    ) as HTMLInputElement).value = JSON.stringify(
```
jbpenrath, line 148:
```suggestion
    const $target = document.getElementById(props.target)
    if ($target) {
        $target.value = JSON.stringify(
```

src/frontend/js/types/Enums.ts
```diff
@@ -11,3 +11,9 @@ export enum filterApiRole {
   NOT_MODERATOR = '!moderator',
   MODERATOR = 'moderator',
 }
+
+export enum FileUploadError {
+  NONE = 0,
+  MAX_UPLOAD = 1,
+  ERROR = 2,
+}
```
jbpenrath, line 19: You could use string. Use integer like you did is implicit (https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums).

In the case of errors, it could useful to use string instead of integer (e.g: if you send to improve readability of errors in Sentry)

```suggestion
export enum FileUploadError {
  NONE = 'NONE',
  MAX_UPLOAD = 'MAX_UPLOAD',
  ERROR = 'ERROR',
}
```

jbpenrath (commented):
First feedbacks about frontend

src/ashley/models.py
madmatah: In fact, necessary or not, I don't understand the purpose of this. That's why a comment would be great!

src/ashley/models.py
carofun: I don't want anything due to saving in database to block the edition or draft-js, saving the image in the database is not primary so blank or null by default seems to me to be advised

src/ashley/models.py
carofun: no purpose i'll get ride of ti

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  render() {
+    const popoverClassName = this.state.open
+      ? 'add-image-popover'
+      : 'add-image-closed-popover';
+    const buttonClassName = this.state.open ? 'add-image-pressed-button' : '';
+    const isErrorOn = this.state.errorOnUpload !== FileUploadError.NONE;
+    const listTypeAllowed = appFrontendContext.typeimage.join(',');
+    return (
+      <div className="bi09khh">
```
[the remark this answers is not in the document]
carofun: it's the class for draft-js toolbar's plugin, i must add a comment about it !

src/ashley/models.py
madmatah: I think we should have a minimal amount of confidence in our data model and our database.
Have you identified a case where this could cause problem ?

carofun: I've always set it this way by principle so I wouldn't know but it's a constraint that i didn't see the point on dealing with, not to care was to me the good level for this task.

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+        <button
+          data-testid="addImage"
+          className={`${buttonClassName} bc4rxid fa fa-image`}
```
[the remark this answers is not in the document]
carofun: ditto it's the class for draft-js toolbar's plugin, i must add a comment about it !

src/frontend/js/components/AshleyEditor/index.tsx
carofun: Thank you for your feedback,  can you explain why $ in the variable's name, is this a convention?
Unfortunately, even with this rewriting, I need to cast it to HTMLInputElement to access the value property.
I'm not sure of the relevance of testing the existence of this target with an if, this dom element is necessary for the code to work so if it doesn't exist nothing will work, IMO a straight error is preferable. WDYT?

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+  modifier(
+    editorState: EditorState,
+    url: string,
+    extraData: Record<string, unknown>,
+  ): EditorState;
+  forum: number;
+  poster: number;
+}
+
+export default class ImageAdd extends React.Component<ImageAddProps> {
```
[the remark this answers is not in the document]
carofun: maybe I could do that in another PR ?

src/frontend/js/components/AshleyEditor/index.tsx
jbpenrath: My mistake. I did not take context into account. 
It's my convention to prefix variables which stores a DOM element with a `$` (jQuery inheritance 😄)

I agree with your comment about the relevance to test the existence of target.

```
const { value } = document.getElementById(props.target) as HTMLInputElement;
```

src/ashley/editor/decorators.py
```diff
@@ -59,3 +59,34 @@ def mention(props):
...
+        css_class = f"float-{alignment}"
+    if alignment == "center":
+        css_class = "image-center"
+
+    # make sure data are number and positive, default is 40% of editor size
+    width = f'{abs(int(props.get("width", "40")))}%'
+
+    height = "auto"
+    if props.get("height", None) is not None:
+        height = f'{abs(int(props.get("height")))}%'
```
madmatah, line 82: Why isn't it done with CSS?
This HTML be stored in database and it will be painful to update if you want to change these values later.

src/ashley/editor/decorators.py
```diff
@@ -59,3 +59,34 @@ def mention(props):
...
+
+    return DOM.create_element(
+        "img",
+        {
+            "class": css_class,
+            "src": props.get("src", "#"),
```
madmatah, line 88: If `props.src` is undefined, should we really return a useless IMG element?

src/ashley/api/views.py
```diff
@@ -90,3 +92,12 @@ def update(self, request, pk=None):
...
         return Response({"success": True}, status=status.HTTP_200_OK)
+
+
+# pylint: disable=too-many-ancestors
+class ImageUploadApiView(viewsets.ModelViewSet):
+    """Viewset for the API of the Image object."""
+
+    queryset = UploadImage.objects.all()
+    serializer_class = UploadImageSerializer
+    permission_classes = [IsAuthenticated]
```
madmatah, line 103: When a `UploadImage` object is created or updated, the  `poster` field should be retrieved from the authenticated user that is doing the request. Or at least you should ensure that it is the same.

We don't want to trust the user input on this one.

src/ashley/models.py
```diff
@@ -197,3 +204,45 @@ class Meta:
...
+    """Abstract model for images."""
+
+    file = models.ImageField(
+        upload_to=image_directory_path,
+        max_length=100,
+        blank=True,
+        validators=[
+            validate_upload_image_file_size,
+            validate_upload_image_allowed_type,
+        ],
```
madmatah, line 226: I'd use [FileExtensionValidator](https://docs.djangoproject.com/en/dev/ref/validators/#fileextensionvalidator) instead of the `validate_upload_image_allowed_type` validator.

src/ashley/validators.py
```diff
@@ -39,3 +41,20 @@ def error(self, message):
...
+        raise ValidationError(
+            f"The maximum file size that can be uploaded is {settings.MAX_UPLOAD_FILE_MB}MB"
+        )
+    return file
+
+
+def validate_upload_image_allowed_type(file):
+    """Gets the extension of the file and check if it is authorized"""
+    ext = (os.path.splitext(file.name)[1])[1:]
+    if f"image/{ext}" not in settings.IMAGE_TYPE_ALLOWED:
```
madmatah, line 58: The setting `IMAGE_TYPE_ALLOWED` is a list of authorized MIME types, which is great.

But it is used here as a file extension checker.
There is not always an exact match between a mime type and the associated file extension.

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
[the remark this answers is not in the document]
jbpenrath: It looks like a generated name. Do we have to update this className at each update of the draft-js toolbar plugin?

src/ashley/editor/decorators.py
carofun: what do you mean done in css ? height and width could be 1 to 100%, I don't see how to deal with this directly in css, could you give me a hint on that please :-)

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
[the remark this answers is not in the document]
jbpenrath: Up to you.
As this component has been added with this PR, from my point of view we could rewrite it right now as a functional component. But you are not using hook in this component, so it is something mandatory right now.

src/ashley/editor/decorators.py
madmatah: You are right, I had misunderstood this code. Sorry you can ignore my comment.

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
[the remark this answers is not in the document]
madmatah: +1 to rewrite it in a functional way in this PR, since that's the way components are written in this project.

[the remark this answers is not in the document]
carofun: ok I misunderstood then, I thought you were telling me that this way of writing component was good and all the other ones should be like this one. Obviously you meant the other way round :-)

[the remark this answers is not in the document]
jbpenrath: Whoops sorry. Nope, with the release of hooks, functional components have become the norm.

src/ashley/machina_extensions/forum_conversation/forms.py
```diff
@@ -5,46 +5,44 @@
...
-
-        list_active_users = [
+        # collect extra informations used by the editor
+        self.fields["content"].widget.attrs.update(
             {
-                "name": get_forum_member_display_name(user),
-                "user": user.pk,
+                "mentions": self.topic.get_active_users(self.user),
+                "forum": self.forum.id,
+                "poster": self.user.id,
```
madmatah, line 22: I think it is unnecessary to give the `poster` prop to the frontend, since we can retrieve it in the DRF view.
I would remove the `poster` prop entirely from the react component.

Another opinion would be appreciated on this point @jbpenrath

carofun: no I had the same thought as well

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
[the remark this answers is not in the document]
carofun: Not to my understanding, css class are declared this way for this plugin, I don't see any correlation between version and classNames
Capture d’écran 2021-06-04 à 10 18 01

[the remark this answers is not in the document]
jbpenrath: After a check, I confirme my assumption. `lib` is the folder where compiled files take place.
This package uses `CSS-in-JS`. A [little script](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/scripts/build-css.js) is in charge to extract CSS from JS files then generate a `plugin.css` file. So at each build, the className may change.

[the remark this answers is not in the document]
carofun: it's a foolish way to do so, there's not even a pattern to target these classes...That's no sense to me, I'll dig into it.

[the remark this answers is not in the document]
carofun: ok I get it ...it can be accessed another way so we don't care about the final name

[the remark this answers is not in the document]
jbpenrath: It's CSS-in-JS. The CSS is strictly linked to its component.

src/frontend/scss/objects/_editor.scss
```diff
@@ -82,3 +82,179 @@
...
+  -webkit-animation: sk-bouncedelay 1.4s infinite ease-in-out both;
+}
+.spinner .bounce1 {
+  animation-delay: -0.32s;
+  -webkit-animation-delay: -0.32s;
+}
+.spinner .bounce2 {
+  animation-delay: -0.16s;
+  -webkit-animation-delay: -0.16s;
+}
```
jbpenrath, line 238:
```scss
.spinner {
  align-items: center;
  display: flex;
  height: 100%;
  justify-content: center;
  width: 100%;

  &__bounce {
    animation: sk-bouncedelay 1.4s infinite ease-in-out both;
    background-color: #333;
    border-radius: 100%;
    display: inline-block;
    height: 12px;
    width: 12px;
    -webkit-animation: sk-bouncedelay 1.4s infinite ease-in-out both;
     
    @for $i from 1 through 2 {
      &:nth-child(#{$i}) {
        animation-delay: -0.32s / $i;
        -webkit-animation-delay: -0.32s / $i;
      }
    }
  }
}
```

jbpenrath (commented):
On my screen (Mac OS, Firefox 89.0), `Charger une image` goes to a new line which is not very aesthetic.

Capture d’écran 2021-06-07 à 18 42 40

carofun:
> On my screen (Mac OS, Firefox 89.0), `Charger une image` goes to a new line which is not very aesthetic.
> 
> Capture d’écran 2021-06-07 à 18 42 40

It was actually because it will have another entry next to it to integrate image from a link. As this is not on yet, I will mess around with the futur style and try to keep that on the side.

src/frontend/scss/objects/_editor.scss
carofun: thank you for this great feedback !

sandbox/settings.py
```diff
@@ -126,9 +126,19 @@ class Base(Configuration):
...
+    # AWS
+    AWS_ACCESS_KEY_ID = values.SecretValue()
+    AWS_S3_CUSTOM_DOMAIN = values.Value()
+    AWS_S3_REGION_NAME = values.Value()
+    AWS_S3_URL_PROTOCOL = values.Value("https")
+    AWS_SECRET_ACCESS_KEY = values.SecretValue()
+    AWS_STORAGE_BUCKET_NAME = values.Value()
+    AWS_DEFAULT_ACL = "public-read"
+    AWS_QUERYSTRING_AUTH = False
+    DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
```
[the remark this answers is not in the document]
madmatah: For the development of this feature, of course you had to use this.

But I agree with @jbpenrath, for newcomers wanting to test / contribute on Ashley, it would be more convenient to use the default storage engine. I see no reason to force them to use S3. My suggestion is to: 
 - use `values.Value("storages.backends.s3boto3.S3Boto3Storage")` in the Base settings
 - use `values.Value("django.core.files.storage.FileSystemStorage")` in the Dev settings. This would allow developers to override this value with an environment variable to work on S3 upload if necessary.

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+    : 'add-image-closed-popover';
+  const buttonClassName = open ? 'add-image-pressed-button' : '';
+  const isErrorOn = errorOnUpload !== FileUploadError.NONE;
+  const listTypeAllowed = appFrontendContext.image_type
+    .map((ext) => '.' + ext)
+    .join(',');
+  const { buttonWrapper, button } = props.theme;
+
+  return (
+    <div className={buttonWrapper!} ref={node}>
```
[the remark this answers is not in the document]
carofun: sweet, nice tip! thank you

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const [errorOnUpload, setErrorOnUpload] = useState(FileUploadError.NONE);
+  const [isImageLoading, setIsImageLoading] = useState(false);
+  const node = useRef() as React.MutableRefObject<HTMLInputElement>;
+
+  const openPopover = () => {
+    if (!open) {
+      setErrorOnUpload(FileUploadError.NONE);
+    }
+    setOpen(!open);
+  };
```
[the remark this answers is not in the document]
carofun: If I click on the button to open the pop-up, I can click on it again to close it. I've added a comment to clarify this point.

src/frontend/js/components/AshleyEditor/ImageAdd/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const evt = event.target as HTMLInputElement;
+    if (evt.files && evt.files!.length > 0) {
+      const file = evt.files[0];
+      if (file.size > appFrontendContext.max_upload * 1024 * 1024) {
+        setErrorOnUpload(FileUploadError.MAX_UPLOAD);
+      } else {
+        uploadImage(evt.files[0]);
+        // if user wants to upload the same image again
+        event.target.value = '';
+      }
```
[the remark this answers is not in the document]
carofun: nice ! thank you !

jbpenrath (approved):
Front end part approved! 🎉 Let's see @madmatah for the rest .

Just an idea of improvement for later: add support of drag'n drop to upload file.

LGTM

sandbox/settings.py
```diff
@@ -126,9 +126,19 @@ class Base(Configuration):
...
     STATICFILES_DIRS = (MACHINA_MAIN_STATIC_DIR,)
 
+    # AWS
+    AWS_ACCESS_KEY_ID = values.SecretValue()
+    AWS_S3_CUSTOM_DOMAIN = values.Value()
+    AWS_S3_REGION_NAME = values.Value()
+    AWS_S3_URL_PROTOCOL = values.Value("https")
+    AWS_SECRET_ACCESS_KEY = values.SecretValue()
+    AWS_STORAGE_BUCKET_NAME = values.Value()
+    AWS_DEFAULT_ACL = "public-read"
```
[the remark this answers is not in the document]
madmatah: Unless i'm wrong, this is not resolved?

tests/ashley/api/test_api_upload_image.py
```diff
@@ -0,0 +1,203 @@
...
+        # Gets last id generated
+        image_db = UploadImage.objects.last()
+
+        # Image should have be added with right values
+        self.assertEqual(UploadImage.objects.count(), 1)
+        self.assertEqual(image_db.forum.id, forum.id)
+        self.assertEqual(image_db.poster.id, user.id)
+
+        # Controls image's path is respected on server
+        self.assertIn(f"image_uploads/{forum.id}/{user.id}/", str(image_db.file))
```
madmatah, line 113:
```suggestion
        self.assertIn(f"image_uploads/{forum.id}/{user.id}/", image_db.file.name)
```

tests/ashley/api/test_api_upload_image.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Request library doesn't know how to handle local files
+        if settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
+            # Controls url returned exists and contains data
+            request = urllib.request.Request(
+                f"http://{settings.AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com/{str(image_db.file)}"
+            )
+            fileHandle = urllib.request.urlopen(request)
+            data = fileHandle.read()
+            self.assertTrue(data is not None)
```
madmatah, line 123: This test only works for the S3Boto3Storage backend.
I propose the following alternative that delegates the retrieval of the file content to the backend storage.
And it checks that the file content is not empty:

```suggestion
        uploaded_file = image_db.file.open()
        self.assertGreater(uploaded_file.size, 0)
        self.assertIsNotNone(uploaded_file.read())
```

This way, it will work with all the storage backends.

src/ashley/models.py
```diff
@@ -197,3 +205,46 @@ class Meta:
...
 LTIContext = model_factory(AbstractLTIContext)
+
+
+def image_directory_path(instance, filename):
+    """File will be uploaded to specific path"""
+    return "image_uploads/{0}/{1}/{2}".format(
+        instance.forum.id,
+        instance.poster.id,
+        f"{uuid.uuid4()}{pathlib.Path(filename).suffix}",
+    )
```
madmatah, line 216: You are mixing string interpolation styles (`.format` vs `f-strings`)
I suggest the following : 

```suggestion
    new_filename = "f"{uuid.uuid4()}{pathlib.Path(filename).suffix}"
    return f"image_uploads/{instance.forum.id}/{instance.poster.id}/{new_filename}"
```

Or maybe a longer, but maybe easier to understand: 

```suggestion
    target_directory = f"image_uploads/{instance.forum.id}/{instance.poster.id}"
    file_extension = pathlib.Path(filename).suffix
    new_filename = "f"{uuid.uuid4()}{file_extension}"
    return f"{target_directory}/{new_filename}"
```

tests/ashley/api/test_api_upload_image.py
```diff
@@ -0,0 +1,203 @@
...
+from machina.core.db.models import get_model
+from PIL import Image
+
+from ashley.factories import ForumFactory, UploadImageFactory, UserFactory
+
+User = get_user_model()
+UploadImage = get_model("ashley", "UploadImage")
+
+
+class ImageUploadApiTest(TestCase):
```
madmatah, line 22: I'd add 3 tests : 

- Check that a user cannot delete any uploaded file
- Check that a user cannot update any uploaded file
- Check that if a user uploads 2 files with the same filename, in the same course, they have distinct urls (you handled this case but i don't see any test to confirm it)

tests/ashley/api/test_api_upload_image.py
```diff
@@ -0,0 +1,203 @@
...
+        self.assertEqual(UploadImage.objects.count(), 0)
+        # Upload image
+        response = self.client.post("/api/v1.0/images/", data, format="multipart")
+        self.assertEqual(response.status_code, 201)
+
+        # Poster id should be the current user id logged in
+        self.assertEqual(UploadImage.objects.count(), 1)
+
+    def test_api_create_upload_bad_extension(self):
+        """Make sure we can't upload a file with unauthorized extension"""
```
madmatah, line 159: ``
```suggestion
        """Make sure we can't upload a file with an unauthorized extension"""
```

madmatah (approved):
...and backend part approved :partying_face: 
Congrats for this great feature!

PR #203, opened 2021-06-14
🗑️(back) filter search on current lti_context

carofun:
## Purpose

A user can be part of multiple forums attached to different classes.
We want the search to keep is current scope and only search in forums
that are related to the course user is actually logged in. Currently,
a user can search in all the forums he has access to and it can be a
bit confusing. Clicking on a result that lead to another course,
connects the user in a different forum from is original search. Results
might not be even relevant as context is different than expected. To be
more relevant and not to loose our users, we filter results to current
LTIContext and only show results from forum of the current class.

## Proposal

Add a filter to only search on forums from this LTIContext to restrict search

### DIFF
--- a/src/ashley/machina_extensions/forum_search/forms.py
+++ b/src/ashley/machina_extensions/forum_search/forms.py
@@ -3,8 +3,12 @@
     ==================
     This module defines forms provided by the ``forum_search`` application.
 """
-
 from machina.apps.forum_search.forms import SearchForm as MachinaSearchForm
+from machina.core.db.models import get_model
+from machina.core.loading import get_class
+
+Forum = get_model("forum", "Forum")
+PermissionHandler = get_class("forum_permission.handler", "PermissionHandler")
 
 
 class SearchForm(MachinaSearchForm):
@@ -12,6 +16,25 @@ class SearchForm(MachinaSearchForm):
     Override Django Machina's SearchForm to allow searching poster user names
     even with an empty search in the main field.
     """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Init form, code based on base class MachinaSearchForm. A filter on lti_context
+        has been added.
+        """
+        # Loads current lti_context to filter forum search
+        lti_contexts = kwargs.pop("lti_context", None)
+        super().__init__(*args, **kwargs)
+        user = kwargs.pop("user", None)
+        self.allowed_forums = PermissionHandler().get_readable_forums(
+            Forum.objects.filter(lti_contexts=lti_contexts), user
+        )
+        # self.allowed_forums is used in search method of MachinaSearchForm
+        if self.allowed_forums:
+            self.fields["search_forums"].choices = [
+                (f.id, "{} {}".format("-" * f.margin_level, f.name))
+                for f in self.allowed_forums
+            ]
 
     def clean(self):
         """
--- a/tests/ashley/test_forum_search.py
+++ b/tests/ashley/test_forum_search.py
@@ -14,7 +14,15 @@ from django.utils import timezone
 from machina.apps.forum_permission.shortcuts import assign_perm
 from machina.core.db.models import get_model
 
-from ashley.factories import ForumFactory, PostFactory, TopicFactory, UserFactory
+from ashley import SESSION_LTI_CONTEXT_ID
+from ashley.factories import (
+    ForumFactory,
+    LTIContextFactory,
+    PostFactory,
+    TopicFactory,
+    UserFactory,
+)
+from ashley.machina_extensions.forum_search.forms import SearchForm
 
 Forum = get_model("forum", "Forum")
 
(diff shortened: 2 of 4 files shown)


src/ashley/machina_extensions/forum_search/forms.py
```diff
@@ -13,6 +17,23 @@ class SearchForm(MachinaSearchForm):
     even with an empty search in the main field.
     """
 
+    def __init__(self, *args, **kwargs):
```
sampaccoud, line 20: missing docstring?

src/ashley/machina_extensions/forum_search/forms.py
```diff
@@ -13,6 +17,23 @@ class SearchForm(MachinaSearchForm):
...
+        # load current lti_context to filter forum search
+        lti_contexts = kwargs.pop("lti_contexts", None)
+        super().__init__(*args, **kwargs)
+        user = kwargs.pop("user", None)
+        self.allowed_forums = PermissionHandler().get_readable_forums(
+            Forum.objects.filter(lti_contexts=lti_contexts), user
+        )
+        if self.allowed_forums:
```
sampaccoud, line 28: So if I understand well, you limit the options proposed to the user in the forum filter select box so that s.he can only see allowed forums. But you are not forcing a filter to only return results within these allow forum are you?

src/ashley/machina_extensions/forum_search/forms.py
```diff
@@ -13,6 +17,23 @@ class SearchForm(MachinaSearchForm):
...
+            self.fields["search_forums"].choices = [
+                (f.id, "{} {}".format("-" * f.margin_level, f.name))
+                for f in self.allowed_forums
+            ]
+        elif self.fields.get("search_forums", None) is not None:
+            # The user cannot view any single forum, the 'search_forums' field can be deleted
```
sampaccoud, line 34: I don't understand why you delete this information. As long as the forum is related to the current context, the user should be allowed to filter on it?

src/ashley/machina_extensions/forum_search/forms.py
carofun: the search method is in MachinaSearchForm class and uses the allowed_forums so it filters on it as well, so in both case it's filtered

```
def search(self):
        sqs = super().search()

        if not self.is_valid():
            return self.no_query_found()

        # Handles topic-based searches
        if self.cleaned_data['search_topics']:
            sqs = sqs.filter(topic_subject=AutoQuery(self.cleaned_data['q']))

        # Handles searches by poster name
        if self.cleaned_data['search_poster_name']:
            sqs = sqs.filter(poster_name__icontains=self.cleaned_data['search_poster_name'])

        # Handles searches in specific forums if necessary
        if 'search_forums' in self.cleaned_data and self.cleaned_data['search_forums']:
            sqs = sqs.filter(forum__in=self.cleaned_data['search_forums'])
        else:
            forum_ids = self.allowed_forums.values_list('id', flat=True)
            sqs = sqs.filter(forum__in=forum_ids) if forum_ids else sqs.none()

        return sqs
```

src/ashley/machina_extensions/forum_search/forms.py
carofun: I kept the original code of the base class MachinaSearchForm but only added my filter.
It deletes the field to know in search method in which case it is
the original code of the base class is this one 
`def __init__(self, *args, **kwargs):
        user = kwargs.pop('user', None)

        super().__init__(*args, **kwargs)

        # Update some fields
        self.fields['q'].label = _('Search for keywords')
        self.fields['q'].widget.attrs['placeholder'] = _('Keywords or phrase')
        self.fields['search_poster_name'].widget.attrs['placeholder'] = _('Poster name')

        self.allowed_forums = PermissionHandler().get_readable_forums(Forum.objects.all(), user)
        if self.allowed_forums:
            self.fields['search_forums'].choices = [
                (f.id, '{} {}'.format('-' * f.margin_level, f.name)) for f in self.allowed_forums
            ]
        else:
            # The user cannot view any single forum, the 'search_forums' field can be deleted
            del self.fields['search_forums']`

carofun: In that case, thinking twice about it, the field is already deleted at this point as we call the init of MachinaSearchForm.

carofun: In that case, thinking twice about it, the field is already deleted at this point as we call the init of MachinaSearchForm.

src/ashley/machina_extensions/forum_search/forms.py
sampaccoud: My understanding is that it filters on a forum if it is passed as parameter. Otherwise you have just limited the filter choices allowed forum but not forced it to the current forum.

Maybe what you should do is override this search method and hardcode a filter on the forums of the current context?

But first, we need to design a test that proves if my concerns are justified : a user allowed on 2 forums and accessing Ashley with a context targetting the first forum, should not see a result posted on the second forum?

carofun: this test is one already unless i misunderstood you

tests/ashley/test_forum_search.py
```diff
@@ -445,3 +452,70 @@ def test_forum_search_empty_public_username(self):
...
+        user = UserFactory()
+
+        lti_context = LTIContextFactory(lti_consumer=user.lti_consumer)
+        lti_context2 = LTIContextFactory(lti_consumer=user.lti_consumer)
+        forum = ForumFactory()
+        forum2 = ForumFactory()
+        forum3 = ForumFactory()
+        forum.lti_contexts.add(lti_context)
+        forum2.lti_contexts.add(lti_context)
+        forum3.lti_contexts.add(lti_context2)
```
sampaccoud, line 468: May be easier to follow if you name your contexts : `context11`, `context12` and `context2`?

src/ashley/machina_extensions/forum_search/forms.py
sampaccoud: Ah ok got it.
So I think you should also test what happens if someone specifies an unauthorized forum in the `search_forums` parameter. It should return a 400 if I understand well?

carofun: And me too I get what you are saying, is that the original code is trusting the search_forums field without checking again if it's part of the allowed forums. I will had a constraint to make this completely clean, so override the search method. 
To allow or not the access on topics that are not part of the lti context and deal with 400 request, that won't be dealt in this PR, this is relative to all the project, not just relative to the search and will be done in a separate PR.

tests/ashley/test_forum_search.py
carofun: oups sorry, don't really get this feedback, what's the difference between context2 and context12 ? In all the other tests of this project, the norm was more to use the variable name lti_context rather than context11.

sampaccoud: nevermind this comment it's a small matter. It's just that naming variable 1, 2, 3 is hard to follow. Giving them meaningful names is easier. The idea was to convey the name the fact that one context is linked to forum1 and others to forum2.

src/ashley/machina_extensions/forum_search/forms.py
sampaccoud: I think it is already taken care of because Django won't accept a field value that is now in choices? I was just asking to add a test to secure it because all these overrides are fragile.

carofun: So I did the extra test, but maybe we misunderstood each other. I don't get a 400 but a 200 with an error message, telling me to choose a good forum because the one chosen is not part of my choices.

sampaccoud: ok! It's interesting that you don't get a 400 :man_shrugging: But as long as you don't access the content it should be fine.

sampaccoud (approved):

PR #204, opened 2021-06-16
🩹(forum) fix sorting on sticky and announcements topics

carofun:
## Purpose

There are different kind of topics: regular, sticky and announcements topics.
Sticky topics are supposed to stay on top of the topic listing. Announcements
are displayed in another block that is above the regular topics block and has
its own header columns. Previous commit 3f92f4a enabled sorting topics on
different header columns. This new feature introduces two different
inconstancies. First, sticky topics were sorted as well whereas they are meant
to stay on top of the listing. Second, columns of announcements block when the
order was requested in URL showed toggling arrow as if they were sorted as well.
We only want to sort regular topic discussions and not announcements.
Announcements must stay with the default order which is ordered on the creation
date. The commit introduces these two fixes.

## Proposal

- order as first order topics by type to keep sticky topics on top
- change the view so the announcement block don't enable to sort header

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - add django rest framework to promote/revoke moderators for current LTIContext
  - add new role moderator and permission to manage moderators
 
+### Fixed
+ - fix sorting on sticky and announcements topics
+
 ## [1.0.0-beta.5] - 2020-03-01
 
 ### Changed
--- a/tests/ashley/machina_extensions/forum/test_forum_view.py
+++ b/tests/ashley/machina_extensions/forum/test_forum_view.py
@@ -1,7 +1,12 @@
+import lxml.html  # nosec
 from django.test import TestCase
+from lxml import etree  # nosec
 from machina.apps.forum_permission.shortcuts import assign_perm
+from machina.core.db.models import get_model
 
 from ashley.factories import ForumFactory, PostFactory, TopicFactory, UserFactory
+
+Topic = get_model("forum_conversation", "Topic")
 
 
 class TestForumView(TestCase):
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - add django rest framework to promote/revoke moderators for current LTIContext
  - add new role moderator and permission to manage moderators
 
+### Fixed
+ - fix sorting on sticky and announcements topics
```
madmatah, line 29: It is not because of this PR, but I see 2 `Added` sections in the Changelog.
Maybe it's an occasion to fix this?

tests/ashley/machina_extensions/forum/test_forum_view.py
```diff
@@ -364,3 +369,177 @@ def test_testing_query_param(self):
...
+        url = f"{url_list_topic}"
+        response = self.client.get(url)
+
+        topicC = forum.topics.get(subject__startswith="TOPIC C the newest")
+        topicB = forum.topics.get(subject__startswith="TOPIC B the eldest")
+        topicA = forum.topics.get(subject__startswith="TOPIC A created second")
+        topicSticky = forum.topics.get(subject__startswith="TOPIC STICKY ONE")
+
+        # Controls topics have been created in the order assumed in their titles
+        self.assertTrue(
```
madmatah, line 393: If this assertion fails, it is hard to debug: pytest will just say something like "assertion is not true" and you will not know which value is incorrect.

I suggest to use `assertGreater` or `assertLess` instead to ease debug.

tests/ashley/machina_extensions/forum/test_forum_view.py
```diff
@@ -364,3 +369,177 @@ def test_testing_query_param(self):
...
+            topic=TopicFactory(forum=forum, type=Topic.TOPIC_STICKY, views_count=7),
+            subject="TOPIC STICKY ONE",
+        )
+
+        topic12 = forum.topics.get(subject__contains="12 views_count")
+        topic9 = forum.topics.get(subject__contains="9 views_count")
+        topic6 = forum.topics.get(subject__contains="6 views_count")
+        topicSticky = forum.topics.get(subject__startswith="TOPIC STICKY ONE")
+        # Confirms that the sticky topic has neither max or lowest view_count
+        self.assertTrue(
```
madmatah, line 433: ditto

tests/ashley/machina_extensions/forum/test_forum_view.py
```diff
@@ -364,3 +369,177 @@ def test_testing_query_param(self):
...
+        self.assertTrue("Replies" in announce_block)
+        self.assertTrue("Views" in announce_block)
+        self.assertTrue("Last post" in announce_block)
+
+        # There's no sortable informations
+        self.assertFalse("sortable sorted" in announce_block)
+        # There's no column that has a sorting link on
+        self.assertFalse("<a href=" in announce_block)
+        # There's no toggle sorting
+        self.assertFalse("Toggle sorting" in announce_block)
```
madmatah, line 488: Using `assertIn` and `assertNotIn` would me more appropriate here.

tests/ashley/machina_extensions/forum/test_forum_view.py
```diff
@@ -364,3 +369,177 @@ def test_testing_query_param(self):
...
+        PostFactory(
+            topic=topicAnnounce1,
+            subject="TOPIC A TYPE ANNOUNCED",
+        )
+        PostFactory(
+            topic=topicAnnounce2,
+            subject="TOPIC B TYPE ANNOUNCED",
+        )
+        # Post of topicAnnounce2 have been created last, it should be the first one on the list
+        self.assertTrue(topicAnnounce1.last_post_on < topicAnnounce2.last_post_on)
```
madmatah, line 515: ditto (`assertLess`)

CHANGELOG.md
carofun: I agree with you, but I already fixed it on another current PR which is almost validated, the pr for the upload

tests/ashley/machina_extensions/forum/test_forum_view.py
carofun: I see what you are saying but that would force me to compare two by two and give more importance than needed for this test.
This test is only a casual test to give context to the reader, we don't have doubt about it, it's only for understanding as we create topics from a sub function. I get the idea of debugging but in this current context, the assertTrue seems more appropriate. WDYT ?

tests/ashley/machina_extensions/forum/test_forum_view.py
carofun: better this way, thanks !

tests/ashley/machina_extensions/forum/test_forum_view.py
carofun: ditto for me too :-)

tests/ashley/machina_extensions/forum/test_forum_view.py
madmatah: I understand that the code will be less pleasant to read, but IMO this is more important to get quickly to the bug if a test is failing.

If a test doesn't bring any additional value and it is proven that it cannot fail, perhaps it has no reason to exist?

Generally, I would avoid using `assertTrue/assertFalse` except if i really want to check a variable containing a boolean value. All tests can be written with these 2 assertions, but it's really painful to debug.

carofun: It gives understanding as explained above, I wouldn't have done this test if I wasn't using a function to create my post and topics, it's for the user that read not to have to worry and give the context directly in one look.
If i break it in 5 lines, imo it will be less obvious that we are only giving context, we are testing below one by one and that's what really matters in this test.
So to me it's not a matter that the code would be less pleasant, it's a matter of relevance, giving 5 lines on that makes it important where it's not, it's just to inform, ok they have been created away in another function but look, they are ordered  the same way they are call and one line for reassurance is good.
That's my last shot, after I'll change it :-)

carofun: Anyway it's not a big deal lol we are only talking about this change

`# Controls topics have been created in the order assumed in their titles
        self.assertGreater(topicSticky.last_post_on,topicC.last_post_on)
        self.assertGreater(topicC.last_post_on,topicA.last_post_on)
        self.assertGreater(topicA.last_post_on,topicB.last_post_on)`

self.assertTrue( topicSticky.last_post_on  > topicC.last_post_on > topicA.last_post_on > topicB.last_post_on )

tests/ashley/machina_extensions/forum/test_forum_view.py
```diff
@@ -364,3 +369,177 @@ def test_testing_query_param(self):
             '<a href="?o=3&whatever=2" class="toggle descending" title="Toggle sorting"></a>',
             html=True,
         )
+
+    def test_testing_topic_is_sticky_stay_sticky_default_order(self):
+        """Request page with a sticky topic and check that the sticky topic is always the first
```
sampaccoud, line 374: We usually put `"""` alone on their lines when the docstring can't fit in a single line. I'm surprised the linter doesn't signal it?

tests/ashley/machina_extensions/forum/test_forum_view.py
```diff
@@ -364,3 +369,177 @@ def test_testing_query_param(self):
...
+        )
+
+        # Calls page with default order
+        url = f"{url_list_topic}"
+        response = self.client.get(url)
+
+        topicC = forum.topics.get(subject__startswith="TOPIC C the newest")
+        topicB = forum.topics.get(subject__startswith="TOPIC B the eldest")
+        topicA = forum.topics.get(subject__startswith="TOPIC A created second")
+        topicSticky = forum.topics.get(subject__startswith="TOPIC STICKY ONE")
```
sampaccoud, line 390: this calls for a traditional use of factories in your method because you could have caught the instances from their creation?

I would do all the models creation in the test function and use randomness on everything else including the title to make sure it's sorting on the date indeed. To illustrate my concern, your topics could be sorted on the title and it would work the same because your titles are C > B > A as well...

carofun: Sam this has been done for the previous pr not to repeat everything every time and as a compromise.
The line below explain that it can be trusted
`self.assertTrue(topicC.last_post_on > topicA.last_post_on > topicB.last_post_on)`

tests/ashley/machina_extensions/forum/test_forum_view.py
```diff
@@ -364,3 +369,177 @@ def test_testing_query_param(self):
...
+        result on default order"""
+        forum, url_list_topic = self._get_url_list_topic_with_three_topics()
+        # Creates a post for sticky topic
+        PostFactory(
+            topic=TopicFactory(forum=forum, type=Topic.TOPIC_STICKY),
+            subject="TOPIC STICKY ONE",
+        )
+
+        # Calls page with default order
+        url = f"{url_list_topic}"
```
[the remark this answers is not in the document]
carofun: no no point, it changed in one line

tests/ashley/machina_extensions/forum/test_forum_view.py
carofun: no lint was happy, it's fixed and noted for next time

madmatah (approved):

sampaccoud (approved):

PR #205, opened 2021-06-17
🔧(sandbox) allowing multiple forums for a same LTIContext on sandbox

carofun:
## Purpose

The sandbox enables to run Ashley in an `iframe`. So far a unique lunch url was set
and its value was defined according to the course id. This was blocking the option
of having multiple forums for the same course using the sandbox. As we are adding
constraints on seeing forums from the same LTIContext, this limit is stopping us from
reproducing cases that we are trying to implement. Adding a lunch URL that can be set
with different values will fix this limit.

## Proposal

Add a field to set up a different lunch URL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - add frontend internationalisation and react components autoload
  - add django rest framework to promote/revoke moderators for current LTIContext
  - add new role moderator and permission to manage moderators
+ - add a `LTI ID` field in the sandbox settings to be able to have multiple
+   forums for a same LTIContext
 
 ### Fixed
  - fix sorting on sticky and announcements topics
--- a/sandbox/dev_tools/forms.py
+++ b/sandbox/dev_tools/forms.py
@@ -1,5 +1,7 @@
 """Forms definition for the dev_tools application."""
 
+import uuid
+
 from django import forms
 from django.utils.translation import gettext_lazy as _
 from lti_toolbox.models import LTIPassport
@@ -32,6 +34,11 @@ class LTIConsumerForm(forms.Form):
     )
     course_title = forms.CharField(
         label="Course Title", max_length=100, initial="Mathematics 101"
+    )
+
+    forum_lti_id = forms.UUIDField(
+        label="LTI ID",
+        initial=uuid.uuid5(uuid.NAMESPACE_DNS, "default"),
     )
 
     role = forms.ChoiceField(
--- a/sandbox/dev_tools/views.py
+++ b/sandbox/dev_tools/views.py
@@ -28,11 +28,7 @@ def dev_consumer(request: HttpRequest) -> HttpResponse:
             launch_url = request.build_absolute_uri(
                 reverse(
                     "forum.lti.view",
-                    kwargs={
-                        "uuid": uuid.uuid5(
-                            uuid.NAMESPACE_DNS, form.cleaned_data["course_id"]
-                        )
-                    },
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - add frontend internationalisation and react components autoload
  - add django rest framework to promote/revoke moderators for current LTIContext
  - add new role moderator and permission to manage moderators
+ - add a lunch url field in the sandbox settings to be able to have multiple
```
madmatah, line 27:
```suggestion
 - add a launch url field in the sandbox settings to be able to have multiple
```

sandbox/dev_tools/forms.py
```diff
@@ -34,6 +34,12 @@ class LTIConsumerForm(forms.Form):
         label="Course Title", max_length=100, initial="Mathematics 101"
     )
 
+    launch_url = forms.CharField(
+        label="launch_url",
+        max_length=100,
+        initial="course-v1:openfun+mathematics101+session01",
+    )
+
```
madmatah, line 42:
```suggestion
    forum_lti_id = forms.UUIDField(
        label="LTI ID",
        initial=uuid.uuid5(uuid.NAMESPACE_DNS, "default"),
    )
```

sandbox/dev_tools/views.py
```diff
@@ -30,7 +30,7 @@ def dev_consumer(request: HttpRequest) -> HttpResponse:
                     "forum.lti.view",
                     kwargs={
                         "uuid": uuid.uuid5(
-                            uuid.NAMESPACE_DNS, form.cleaned_data["course_id"]
+                            uuid.NAMESPACE_DNS, form.cleaned_data["launch_url"]
```
madmatah, line 33:
```python
"uuid": form.cleaned_data["forum_lti_id"]
```

madmatah (commented):
Here is a little suggestion that makes more sense to me.

Reusing the course id might be confusing.
You might as well display a UUID that will always be the same by default. And let the possibility to modify it via this field.

I have a few comments about commit message:

- The sandbox enables to run Ashley in an iframe => In development mode, the sandbox allows to use ashley via a dummy LTI consumer.
- lunch url -> launch url (2 times)
- Maybe you can simplify it to "In development mode, the sandbox allows to use ashley via a dummy LTI consumer. It does not allow to create several forums for the same course, since the forum LTI id is generated from the course ID. This prevents testing some ashley features."

carofun:
> * Maybe you can simplify it to "In development mode, the sandbox allows to use ashley via a dummy LTI consumer. It does not allow to create several forums for the same course, since the forum LTI id is generated from the course ID. This prevents testing some ashley features." => Thanks for your feedback, I kind of did a mix between the both!

sandbox/dev_tools/views.py
madmatah: The `uuid.uuid5()` call is no longer necessary here. See my suggestion :arrow_up:

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
  - add frontend internationalisation and react components autoload
  - add django rest framework to promote/revoke moderators for current LTIContext
  - add new role moderator and permission to manage moderators
+ - add a launch url field in the sandbox settings to be able to have multiple
```
madmatah, line 27:
```suggestion
 - add a `LTI ID` field in the sandbox settings to be able to have multiple
```

madmatah (approved):

PR #211, opened 2021-07-13
✨(back) allow instructors to archive a forum

madmatah:
## Purpose

If a forum is created by mistake by an instructor on Ashley, it cannot be deleted. And as a result, it is displayed to all students coming from the same LTI context.

## Proposal

This commit introduces a "delete" action on the forum. It is rather an archiving of the forum than a deletion: the forum is flagged as archived and not accessible anymore.

archive


src/ashley/machina_extensions/forum/views.py
```diff
@@ -51,6 +53,13 @@ def get(self, request, **kwargs):
         response = super().get(request, **kwargs)
         return response
 
+    def get_forum(self):
+        """Returns the forum to consider and check that it has not been archived"""
```
carofun, line 57:
```suggestion
        """Returns the forum to consider and checks that it has not been archived"""
```

src/ashley/machina_extensions/forum/views.py
```diff
@@ -132,6 +141,29 @@ def get_ordering_column(self):
         return self.get_default_index_order().get("col")
 
 
+class ForumArchiveView(PermissionRequiredMixin, UpdateView):
+    """Display the form to archive a forum."""
```
carofun, line 145:
```suggestion
    """Displays the form to archive a forum."""
```

src/ashley/machina_extensions/forum/views.py
```diff
@@ -132,6 +141,29 @@ def get_ordering_column(self):
...
+        """Returns the success URL to redirect the user to."""
+        return reverse("forum:index")
+
+    # pylint: disable=unused-argument
+    def perform_permissions_check(self, user, obj, perms):
+        """Performs the permissions check."""
+        return self.request.forum_permission_handler.can_archive_forum(obj, user)
+
+    def get_form_kwargs(self):
+        """Ignore data posted by the user and force the archiving of the forum"""
```
carofun, line 161:
```suggestion
        """Ignores data posted by the user and forces the archiving of the forum"""
```

tests/ashley/machina_extensions/forum/test_forum_archive.py
```diff
@@ -0,0 +1,80 @@
...
+        self.assertFalse(forum.archived)
+
+        self.client.force_login(user, "ashley.auth.backend.LTIBackend")
+
+        # The user can read the forum
+        response = self.client.get(f"/forum/forum/{forum.slug}-{forum.pk}/")
+        self.assertEqual(200, response.status_code)
+        self.assertContains(response, forum.name)
+
+        # but he's not allowed archive it
```
carofun, line 36:
```suggestion
        # but he's not allowed to archive it
```

tests/ashley/machina_extensions/forum/test_forum_archive.py
```diff
@@ -0,0 +1,80 @@
...
+        forum.lti_contexts.add(lti_context)
+
+        assign_perm("can_read_forum", user, forum, True)
+        assign_perm("can_archive_forum", user, forum, True)
+
+        self.assertFalse(forum.archived)
+
+        self.client.force_login(user, "ashley.auth.backend.LTIBackend")
+
+        # The user can access to the forum
```
carofun, line 63:
```suggestion
        # The user can access the forum
```

tests/ashley/machina_extensions/forum/test_forum_archive.py
```diff
@@ -0,0 +1,80 @@
...
+        response = self.client.get(f"/forum/admin/archive/{forum.pk}/")
+        self.assertEqual(200, response.status_code)
+        self.assertContains(response, "Delete the forum")
+
+        # He can archive it
+        update_response = self.client.post(f"/forum/admin/archive/{forum.pk}/")
+        self.assertEqual(302, update_response.status_code)
+        self.assertTrue(Forum.objects.get(pk=forum.pk).archived)
+
+        # And once it's done, he cannot access to it anymore
```
carofun, line 78:
```suggestion
        # And once it's done, he cannot have access to it anymore
```

carofun (commented):
Great ! it will be a nice feature !!

src/ashley/machina_extensions/forum_permission/handler.py
```diff
@@ -41,5 +45,7 @@ def forum_list_filter(self, qs, user):
         """
         forums_to_show = super().forum_list_filter(qs, user)
```
carofun, line 46: Extra feedback, It makes me wonder if there is no high-level method that would need to be overridden in Django's Machina to be sure deleted forums don't appear. I'm thinking for instance of _get_all_forums from Django's Machina. Maybe it would be a good idea to add a filter directly on a high-level method to prevent errors.
see machina/apps/forum_permission/handler.py

madmatah: Thanks for your feedback!

I don't think we should set this behavior in `_get_all_forums()` because the purpose of this function is to return all the forums, without worrying about permissions, with a caching mechanism. What if we want administrators to be able to see archived forums, for example? It would not be possible if we introduce this behavior in this method.

In addition to that, this function is not used everywhere, there is no high-level method used everywhere.
Multiple django-machina components are directly using `Forum.objects.all()` and apply the `forum_list_filter` function on the result. That's why I chose to override this function.

I agree with you: this could be better. But it implies overriding a lot of django machina components to introduce a high-level method to handle this :/

carofun (approved):
Great, CP is waiting for this !

PR #213, opened 2021-08-12
🐛(back) limit the scope of the unread topics view

madmatah:
## Purpose

When a user displays the list of its unread topics, the list contains
topics belonging to every forums the user has access to. They should
be limited to forums related to the LTIContext the user is coming
from.

## Proposal

- [x] Update the query in machina's permission handler function `get_readable_forums`

### DIFF
--- a/src/ashley/machina_extensions/forum_permission/handler.py
+++ b/src/ashley/machina_extensions/forum_permission/handler.py
@@ -4,6 +4,7 @@ implement filter or access logic related to forums.
 """
 from typing import Optional
 
+from django.db import models
 from machina.apps.forum_permission.handler import (
     PermissionHandler as BasePermissionHandler,
 )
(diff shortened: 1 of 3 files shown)


tests/ashley/machina_extensions/forum_permission/test_permission_handler.py
```diff
@@ -66,3 +71,58 @@ def test_forum_list(self):
...
+        # When the permission handler has no lti context specified,
+        # the get_readable_forums should return all forums the user
+        # has access to
+        forums = Forum.objects.all()
+        readable_forums = permission_handler.get_readable_forums(forums, user)
+
+        self.assertEqual(len(readable_forums), 3)
+        self.assertIn(forum_a1, readable_forums)
+        self.assertIn(forum_a2, readable_forums)
+        self.assertIn(forum_b1, readable_forums)
```
sampaccoud, line 115: I would say `self.assertCountEqual` can do that in one line

tests/ashley/machina_extensions/forum_permission/test_permission_handler.py
```diff
@@ -66,3 +71,58 @@ def test_forum_list(self):
         self.assertNotContains(response, "Forum A2")
         self.assertContains(response, "Forum B1")
         self.assertNotContains(response, "Forum B2")
+
+    def test_get_readable_forums(self):
```
sampaccoud, line 75: I think you also need a test for the case when it is not a queryset but a list?

tests/ashley/machina_extensions/forum_permission/test_permission_handler.py
madmatah: I forgot this one, thanks!

tests/ashley/machina_extensions/forum_permission/test_permission_handler.py
madmatah: Yep, I added it... and it was worth it :grimacing:

src/ashley/machina_extensions/forum_permission/handler.py
```diff
@@ -37,9 +38,32 @@ def forum_list_filter(self, qs, user):
...
+        current LTI context of the User, if any.
         """
         forums_to_show = super().forum_list_filter(qs, user)
         if self.current_lti_context_id:
             return forums_to_show.filter(lti_contexts__id=self.current_lti_context_id)
         return forums_to_show
+
+    def get_readable_forums(self, forums, user):
+        """
+        Returns a queryset of forums that can be read by the considered user.
```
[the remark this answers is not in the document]
sampaccoud: Then the docstring should mention it?

sampaccoud (approved):

src/ashley/machina_extensions/forum_permission/handler.py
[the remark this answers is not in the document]
madmatah: I updated the docstring:

> Given a QuerySet (or list) of forums, it returns the subset of forums that can be read by the considered user, with the same type.
>
>We override django machina's method to filter forums based on the current LTI context of the User, if any.

PR #218, opened 2021-09-14
✨(xapi) track topic views with XAPI events

madmatah:
## Purpose

We want to track our users' actions using the XAPI standard and
collect these events for analysis.

## Proposal

- [x] add `xapi` utility module to help building XAPI statements, using tincan library
- [x] track "topic view" event using Django signals
- [x] Add gelf logger in the sandbox settings as an example

### DIFF
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -317,14 +317,23 @@ class Development(Base):
             "verbose": {
                 "format": "[%(levelname)s] [%(asctime)s] [%(module)s] "
                 "%(process)d %(thread)d %(message)s"
-            }
+            },
+            "gelf": {
+                "()": "logging_gelf.formatters.GELFFormatter",
+                "null_character": True,
+            },
         },
         "handlers": {
             "console": {
                 "level": "DEBUG",
                 "class": "logging.StreamHandler",
                 "formatter": "verbose",
-            }
+            },
+            "gelf": {
+                "class": "logging.StreamHandler",
+                "stream": "ext://sys.stdout",
+                "formatter": "gelf",
+            },
         },
         "loggers": {
             "oauthlib": {"handlers": ["console"], "level": "DEBUG", "propagate": True},
@@ -335,6 +344,9 @@ class Development(Base):
                 "propagate": True,
             },
             "django": {"handlers": ["console"], "level": "INFO", "propagate": True},
+            # This formatter is here as an example to what is possible to do
+            # with xapi loggers.
+            "xapi": {"handlers": ["gelf"], "level": "INFO", "propagate": True},
         },
     }
 
--- a/src/ashley/receivers.py
+++ b/src/ashley/receivers.py
@@ -1,0 +1,105 @@
+"""This module defines signal receivers for the Ashley application"""
+
+import logging
+
+from django.conf import settings
+from django.dispatch import receiver
+from django.utils.translation import to_locale
+from machina.apps.forum_conversation.signals import topic_viewed
+from machina.conf import settings as machina_settings
+from machina.core.db.models import get_model
+from tincan import (
+    Activity,
+    ActivityDefinition,
+    Context,
+    ContextActivities,
+    LanguageMap,
+    Verb,
+)
+
+from .xapi import build_statement
+
+LTIContext = get_model("ashley", "LTIContext")
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(topic_viewed)
+# pylint: disable=unused-argument
+def track_topic_view(sender, topic, user, request, response, **kwargs):
+    """ Log a XAPI statement when a user views a topic. """
+
+    consumer = getattr(user, "lti_consumer", None)
+    if consumer is None:
+        logger.warning("Unable to get LTI consumer of user %s", user)
+        return
+    xapi_logger = logging.getLogger(f"xapi.{user.lti_consumer.slug}")
+
+    verb = Verb(
+        id="http://id.tincanapi.com/verb/viewed",
+        display=LanguageMap({"en-US": "viewed"}),
+    )
+
+    obj = Activity(
+        id=f"id://ashley/topic/{topic.pk}",
+        definition=ActivityDefinition(
+            name=LanguageMap(
+                {to_locale(settings.LANGUAGE_CODE).replace("_", "-"): topic.subject}
+            ),
+            type="http://id.tincanapi.com/activitytype/discussion",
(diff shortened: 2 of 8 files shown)


src/ashley/receivers.py
```diff
@@ -0,0 +1,85 @@
...
+        display=LanguageMap({"en-US": "viewed"}),
+    )
+
+    activity_extensions = {}
+    if request.forum_permission_handler.current_lti_context_id is not None:
+        try:
+            lti_context = LTIContext.objects.get(
+                pk=request.forum_permission_handler.current_lti_context_id
+            )
+            activity_extensions = get_course_extensions(lti_context)
```
quitterie-lcs, line 48: To avoid confusion, I would name it `activity_definition_extensions` or `object_extensions`.
```suggestion
            activity_definition_extensions = get_course_extensions(lti_context)
```
OR
```suggestion
            object_extensions = get_course_extensions(lti_context)
```

src/ashley/receivers.py
```diff
@@ -0,0 +1,85 @@
...
+            pass
+
+    obj = Activity(
+        id=f"id://ashley/topic/{topic.pk}",
+        definition=ActivityDefinition(
+            name=LanguageMap(
+                {to_locale(settings.LANGUAGE_CODE).replace("_", "-"): topic.subject}
+            ),
+            type="http://id.tincanapi.com/activitytype/discussion",
+            extensions=activity_extensions,
```
quitterie-lcs, line 59:
```suggestion
            extensions=activity_definition_extensions,
```
OR
```suggestion
            extensions=object_extensions,
```

quitterie-lcs (commented):
Only a few suggestion

tests/ashley/test_models_lticontext.py
```diff
@@ -271,3 +271,17 @@ def test_get_user_roles_other_context(self):
...
+        id of the context, when it is in the OpenEdX format.
+        """
+
+        context1 = LTIContextFactory(
+            lti_id="course-v1:myschool+mathematics101+session01"
+        )
+        course_info1 = context1.get_course_info()
+        self.assertEqual("myschool", course_info1["school_name"])
+        self.assertEqual("mathematics101", course_info1["course_name"])
+        self.assertEqual("session01", course_info1["course_run"])
```
jmaupetit, line 287: It might be good to also test non-openedx or partial patterns?

src/ashley/models.py
```diff
@@ -189,6 +190,30 @@ def get_user_roles(self, user) -> List[str]:
...
+        # Extract the course informations if the LTI id of this context is in
+        # the OpenEdx format
+        openedx_match = re.match(r"^course-v[0-9]:(.*)$", self.lti_id)
+        if openedx_match:
+            part = openedx_match.group(1).split("+")
+            length = len(part)
+            return {
+                "school_name": part[0] if length >= 1 else None,
+                "course_name": part[1] if length >= 2 else None,
+                "course_run": part[2] if length >= 3 else None,
```
jmaupetit, line 213: Shouldn't we raise an error if not all fields are filled?

jmaupetit (commented):
Looks amazing! :pray:

src/ashley/receivers.py
madmatah: Thanks! I have opted for `activity_definition_extensions`

src/ashley/models.py
madmatah: To be honest, I used the same logic than in Marsha here.

In this case, I'm not comfortable with raising an error when the outside world does not reflect our expectations.
I see it as a "best-effort" data extraction function.

tests/ashley/test_models_lticontext.py
madmatah: Yes, good suggestion, thanks!

sandbox/settings.py
```diff
@@ -335,6 +344,9 @@ class Development(Base):
                 "propagate": True,
             },
             "django": {"handlers": ["console"], "level": "INFO", "propagate": True},
+            # This formatter is here as an example to what is possible to do
+            # with xapi loggers.
+            "xapi": {"handlers": ["gelf"], "level": "INFO", "propagate": True},
```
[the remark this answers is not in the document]
madmatah: I've set the `INFO` level because the xapi logger is currently used with the info level only, and this is just an example.
I can change it if you think it's better with `DEBUG`, I confess have no particular opinion on this point ^^

madmatah:
@quitterie-lcs @jmaupetit Thank you very much for your review, I added a few fixup commits to address them!

sandbox/settings.py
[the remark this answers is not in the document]
quitterie-lcs: Alright, so let's keep `INFO` level :wink:

quitterie-lcs (approved):
LGTM. I would say that's OK for the first event ! :tada:

src/ashley/models.py
jmaupetit: IMO we should be pedantic on this: I would prefer raising an error and fix it if data is missing than emit (and store) a partial/corrupted record (or thousands of them) that will never be used.

The same paradigm applies for Marsha. Ping @lunika :wink:

quitterie-lcs: As discussed, these information could be contained in a parent in the context as followed.

```
context_activities=ContextActivities(
      parent=[
          Activity(
              id=<course_key>,
              definition=ActivityDefinition(
                  name=LanguageMap({"en-US": "test course"}),
                  type="http://adlnet.gov/expapi/activities/course",
              ),
          )
      ]
  )
```
The triplet {school, course, session} would be gathered in the `course_key` (ID of the parent)

madmatah: Thanks! I added a fixup commit to address this.

Note that the course name is not included, we don't store it in the current state of the project.
(And, for example, OpenEdx dogwood does not send it in the LTI request)

src/ashley/receivers.py
```diff
@@ -0,0 +1,105 @@
...
+                Activity(
+                    id=lti_context.lti_id,
+                    definition=ActivityDefinition(
+                        type="http://adlnet.gov/expapi/activities/course"
+                    ),
+                )
+            )
+
+        except LTIContext.DoesNotExist:
+            pass
```
jmaupetit, line 92: Can you add a comment explaining how this can occur?

jmaupetit (approved):
One single suggestion, else:

PR #219, opened 2021-09-16
✨(admin) allow to search users by username in django admin

madmatah:
## Purpose

We want to be able to search users by username in django admin. In its
current state, the User admin view is unusable with thousands of
users.

## Proposal

- [x] Add the `username` field to the search fields in the UserAdmin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- allow to search users by any of its meaningful fields in Django admin
+
 ## [1.0.0] - 2021-08-16
 
 ### Fixed
--- a/src/ashley/admin.py
+++ b/src/ashley/admin.py
@@ -7,3 +7,12 @@ from .models import User
 @admin.register(User)
 class UserAdmin(admin.ModelAdmin):
     """Admin class for Ashley's User model."""
+
+    search_fields = (
+        "username",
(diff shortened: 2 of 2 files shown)


src/ashley/admin.py
```diff
@@ -7,3 +7,5 @@
 @admin.register(User)
 class UserAdmin(admin.ModelAdmin):
     """Admin class for Ashley's User model."""
+
+    search_fields = ["username"]
```
sampaccoud, line 11: Why not add also email, uid, etc while you're at it

carofun (approved):
Great move !!!

Small question? Do we need to write django instead of Django with a capital letter in your commit message

src/ashley/admin.py
madmatah: You're right.
I just added the same fields than in Django's `UserAdmin` + the one that are specific to Ashley.

madmatah:
> Do we need to write django instead of Django with a capital letter in your commit message

@carofun Yep, I fixed it

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- allow to search users by username in Django admin
```
sampaccoud, line 13:
```suggestion
- allow to search users by any of its meaningful fields in Django admin
```

sampaccoud (approved):

PR #223, opened 2021-10-21
✨(auth) set a default public username if empty for admin/instructor

carofun:
## Purpose

When the public username is not provided in the LTI payload, we ask
the user to define the username he would like to use. In some
contexts (i.e. when Ashley is integrated into the Open edX studio),
the public username will then be set for the entire team. In the
external tool loading Ashley, this is confusing. We have decided for
users that have administrator or instructor roles, to force a default
username depending on the role. This will avoid the situation when an
instructor or an administrator was thinking he was setting a username
only for himself.

## Proposal

Description...

- add default public username if empty on authentification and if roles are in administrator or instructor


tests/ashley/test_auth_backend.py
```diff
@@ -379,3 +379,105 @@ def test_moodle_launch_request(self):
...
+            {
+                "user_id": "3fd0ff83-a62d-4a12-9716-4d48821ae24f",
+                "lti_message_type": "basic-lti-launch-request",
+                "lti_version": "LTI-1p0",
+                "resource_link_id": "aaa",
+                "context_id": "course-v1:fooschool+authbackend+0001",
+                "lis_person_contact_email_primary": "user_without_username@example.com",
+                "roles": (
+                    "Administrator,Instructor,urn:lti:sysrole:ims/lis/Administrator,"
+                    + "urn:lti:instrole:ims/lis/Administrator"
```
sampaccoud, line 471:
```suggestion
                    "urn:lti:instrole:ims/lis/Administrator"
```

PR #229, opened 2021-11-03
🐛(back) exclude archived forums from select forms

carofun:
## Purpose

If a forum gets archived, we shouldn't be able to move topics to
this forum. Lists must ignore archived forums. In the same way, if
we want to use the advanced search, we shouldn't be able to search
in a forum that has been archived.

## Proposal

Exclude archived forums from the two listing

### DIFF
--- a/tests/ashley/machina_extensions/forum_moderation/test_forum_moderation.py
+++ b/tests/ashley/machina_extensions/forum_moderation/test_forum_moderation.py
@@ -182,3 +182,94 @@ class ForumModerationTestViewForm(TestCase):
             f"Select a valid choice. {forum3Lti2.id} is not one of the available choices.",
             html=True,
         )
+
+    def test_forum_moderation_list_forums_with_deleted_form(self):
+        """
+        Create three forums in the same LTIContext and archive one. Load the form to choose
+        to which forum the user wants to move the topic. Control that the forum deleted is
+        not part of the choice.
+        """
+        user = UserFactory()
+
+        lti_context = LTIContextFactory(lti_consumer=user.lti_consumer)
+        forum1 = ForumFactory(name="Forum1")
+        forum2 = ForumFactory(name="Forum2")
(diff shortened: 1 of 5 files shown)


tests/ashley/machina_extensions/forum_moderation/test_forum_moderation.py
```diff
@@ -182,3 +182,92 @@ def test_forum_moderation_list_forums_form(self):
...
+
+        # We try to move topic from forum1 to forum2 even if forum2 has been archived
+        response = self.client.post(
+            f"/forum/moderation/topic/{topicForum1.slug}-{topicForum1.id}/move/",
+            data={"forum": {forum2.id}},
+            follow=True,
+        )
+
+        # Control that we get an error and the move is not executed
+        self.assertContains(
```
sampaccoud, line 269: check that error code is 400?

tests/ashley/machina_extensions/forum_moderation/test_forum_moderation.py
```diff
@@ -182,3 +182,92 @@ def test_forum_moderation_list_forums_form(self):
             f"Select a valid choice. {forum3Lti2.id} is not one of the available choices.",
             html=True,
         )
+
+    def test_forum_moderation_list_forums_with_deleted_form(self):
+        """
+        Create three forums in the same LTIContext and archive one. Load the form to choose
+        to which forum the user wants to move the topic. Control that the forum deleted is
+        not part of the choice.
+        """
```
sampaccoud, line 191: add to the docstring that you also test that we get a 400 if the value is submitted.

tests/ashley/machina_extensions/forum_moderation/test_forum_moderation.py
carofun: it's actually a 200, I've added it

sampaccoud (approved):

PR #234, opened 2021-12-31
✨(forum) use the name of a previous forum with the same `lti_id`

carofun:
## Purpose

When duplicating courses from an LMS, forums get copied. By default,
each forum has a name set with the value of the course's name. It's
more likely, when a course gets copied, that instructors want to keep
the same forum's names. To simplify this use case, we define the
default name of each forum with an existing UUID to the previous name
defined for the other `context.id`.

## Proposal

When creating a forum, we search if this lti_id already exists to reuse the name.

### DIFF
--- a/src/ashley/views.py
+++ b/src/ashley/views.py
@@ -64,10 +64,22 @@ class ForumLTIView(BaseLTIAuthView):
                 lti_contexts__id=context.id,
             )
         except Forum.DoesNotExist:
+
+            # Check if this uuid exists for another context.id to reuse its name as a
+            # default value (enables copy and paste)
+            forum_same_uuid = (
+                Forum.objects.only("name")
+                .filter(lti_id=forum_uuid, type=Forum.FORUM_POST)
+                .order_by("-created")
+                .first()
+            )
+
             forum = Forum.objects.create(
                 lti_id=forum_uuid,
                 type=Forum.FORUM_POST,
-                name=lti_request.context_title,
+                name=forum_same_uuid.name
+                if forum_same_uuid
+                else lti_request.context_title,
             )
             logger.debug("Forum %s created", forum_uuid)
             self._init_forum(forum, context)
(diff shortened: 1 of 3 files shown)


src/ashley/views.py
```diff
@@ -64,10 +64,18 @@ def _do_on_login(self, lti_request: LTI) -> HttpResponse:
                 lti_contexts__id=context.id,
             )
         except Forum.DoesNotExist:
+
+            # Check if this uuid exists for another context.id to reuse its name as a
+            # default value (enables copy and paste)
+            forum_same_uuid = Forum.objects.filter(
+                lti_id=forum_uuid, type=Forum.FORUM_POST
```
sampaccoud, line 71: aren't we mixing uuid and lti_id? In line 48 we set `lti_id=context_id` which seems wrong? There is a `lti_id` field in the request that doesn't change in OpenEdX accross course copies.

sampaccoud: you can add `only("name")` since that's the only field you're interested in here.

carofun: The field lti_id in the forum table is the uuid of the forum ex 916180a7-0b43-5c08-b3c8-c738826880bb, the field lti_id in the LTIContext model is the id of the course (l48) example course-v1:openfun+mathematics101+session01.
L48 and L71 are different data, first one is context_id = lti_request.get_param("context_id"), and second is self.kwargs["uuid"]

sampaccoud, line 71:
```suggestion
            forum_same_uuid = Forum.objects.only("name").filter(
                lti_id=forum_uuid, type=Forum.FORUM_POST
```

sampaccoud: as discussed IRL I understood that lti_id is in one cas the LTI id for the context and in the other case the LTI id for the forum object.

sampaccoud (approved):

Issue #233, opened 2021-12-07
Allow sorting forums on the home page


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
On the home page, forums are sorted by their creation date and it is not possible to reorder them.

**Describe the solution you'd like**
Instructors should be able to reorder forums on the home page.

**Describe alternatives you've considered**
I see 2 possible solutions:
1- allow choosing a column by which the forums are sorted
2- allow reordering forums by drag and drop.

The first solution may be easier and good enough but we recently implemented a sortable many-to-many in [joanie](https://github.com/openfun/joanie/pull/64) that could also be a solution depending on how forums are linked to their context.

PR #235, opened 2022-01-05
✨(forum) Allow sorting forums on the home page

carofun:
## Purpose

Courses can have multiple forums and the current order is not
relevant. The view listing the different forums is by default
ordered by creation, the first forum created is shown first.
It’s interesting to be able to sort forums from other parameters
like alphabetical order, number of topics, number of messages,
or date of last message answered.

Resolve https://github.com/openfun/ashley/issues/233

## Proposal

Order by columns

https://user-images.githubusercontent.com/76939263/148242759-435be2ab-c208-4d4c-acf6-506891fc6358.mov


sampaccoud (commented):
Awesome feature and well executed :star_struck: 

I have proposed some simplification on the Mixin stuff :wink: 
I'm available if you want to discuss some patterns.

carofun:
> Awesome feature and well executed 🤩
> 
> I have proposed some simplification on the Mixin stuff 😉 I'm available if you want to discuss some patterns.

Cool many thank for your precious technical return

src/ashley/machina_extensions/forum/views.py
```diff
@@ -5,34 +5,52 @@
...
+        self.params = {}
+
+    def get(self, request, *args, **kwargs):
+        """Handles GET requests."""
+        # collect the current params for the order by in request
+        self.params = dict(request.GET.items())
+        try:
+            super_get = getattr(super(), "get")
+            return super_get(request, *args, **kwargs)
+        except Exception as exception:
```
sampaccoud, line 48: I think you should catch only the precise exception you're expecting

sampaccoud (approved):

src/ashley/machina_extensions/forum/views.py
carofun: j'ai mis except AttributeError as exception:

PR #238, opened 2022-01-11
🐛(auth) set a default public username for existing users

carofun:
## Purpose

In previous commit 06bfbc3, we defined a default public username for
instructors and administrators that had none defined (i.e. when Ashley
is integrated into the Open edX studio). No change had been done
for users that were already set and had an empty public_username, this is now
implemented.

## Proposal

Detect when user exists if he has an empty public_username to reset it to a default value


tests/ashley/test_auth_backend.py
```diff
@@ -482,3 +482,93 @@ def test_automatic_default_public_username_role_administrator_instructor(self):
...
+
+        # No new user have been created
+        self.assertEqual(1, get_user_model().objects.count())
+        # Confirm that public_username is reset to the default value
+        self.assertEqual(public_username, user1.public_username)
+
+    def test_openedx_studio_launch_request_existing_user_all_roles_empty_username(self):
+        """
+        Check that users previously created that have a public_username defined don't get
+        their public_username reset to the next connection.
```
sampaccoud, line 535:
```suggestion
        their public_username reset with the next connection.
```

sampaccoud (approved):

PR #248, opened 2022-03-07
Refacto/promote moderators

carofun:
## Purpose

The endpoint that lists users that are part of a group was
calculated for each user the list of the role he was
part of. This implied extra queries for each user of the
list. Ashley can be used with forums that have a lot of
participants. Adding systematically this information on
querying listing is not necessary and consumes resources
that can be saved.
We create two new endpoints so that we can directly target adding
or removing the group moderator for a user.

## Proposal

- stop calculating roles of users in the serializer
- create two new endpoints add_group_moderator, remove_group_moderator

### DIFF
--- a/src/ashley/api/views.py
+++ b/src/ashley/api/views.py
@@ -2,18 +2,22 @@
 import logging
 
 from django.contrib.auth import get_user_model
-from django.core.exceptions import PermissionDenied
 from machina.apps.forum_permission.viewmixins import (
     PermissionRequiredMixin as BasePermissionRequiredMixin,
 )
 from machina.core.db.models import get_model
 from machina.core.loading import get_class
-from rest_framework import mixins, status, viewsets
+from rest_framework import mixins, viewsets
+from rest_framework.decorators import action
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
 from ashley.context_mixins import get_current_lti_session
-from ashley.defaults import _FORUM_ROLE_INSTRUCTOR, _FORUM_ROLE_MODERATOR
+from ashley.defaults import (
+    _FORUM_ROLE_INSTRUCTOR,
+    _FORUM_ROLE_MODERATOR,
+    _FORUM_ROLE_STUDENT,
+)
 from ashley.permissions import ManageModeratorPermission
 
 from .serializers import UploadImageSerializer, UserSerializer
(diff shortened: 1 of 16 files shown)


src/ashley/api/views.py
```diff
@@ -60,38 +60,52 @@ def get_queryset(self):
...
-                roles,
-                user_groups,
-            )
-            raise PermissionDenied()
+        if _FORUM_ROLE_MODERATOR in user_groups and len(user_groups) > 0:
+            return Response({"detail": "Group moderator can't be added."}, status=400)
 
-        return Response({"success": True}, status=status.HTTP_200_OK)
+        user.groups.add(lti_context.get_role_group(_FORUM_ROLE_MODERATOR))
+        user.save()
```
sampaccoud, line 107: I would say you don't need to save?

sampaccoud, line 107: I would say you don't need to save?

src/frontend/js/data/fetchApi.ts
```diff
@@ -31,17 +31,17 @@ export const fetchUsers = async (role: filterApiRole) => {
 
 /**
  * Request API to add group moderator for this student
+ *
  */
-export const updateUser = async (user: User) => {
+export const updateUserGroup = async (user: User, action: Actions) => {
```
sampaccoud, line 36: or `handleUserModeratorGroup`?

tests/ashley/api/test_api_manage_moderators.py
```diff
@@ -211,13 +212,184 @@ def test_access_can_manage_moderators_list_non_moderators(self):
...
+        forum = ForumFactory()
+        forum.lti_contexts.add(lti_context)
+        lti_context.sync_user_groups(user, ["instructor"])
+        assign_perm("can_manage_moderator", user, forum, True)
+        # Creates the session
+        self.client.force_login(user, "ashley.auth.backend.LTIBackend")
+        session = self.client.session
+        session[SESSION_LTI_CONTEXT_ID] = lti_context.id
+        session.save()
+        response = self.client.post(
```
sampaccoud, line 257: Note that POST should be on the list endpoint. That's why you get a 405. 
But ok you may want to confirm that this endpoint does not accept POST requests.

tests/ashley/api/test_api_manage_moderators.py
```diff
@@ -211,13 +212,184 @@ def test_access_can_manage_moderators_list_non_moderators(self):
...
+        response = self.client.put(
+            f"/api/v1.0/users/{user.id}/",
+            json.dumps(data),
+            content_type="application/json",
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), {"id": 1, "public_username": "user"})
+
+    def test_access_api_moderator_group_anonymous_patch(self):
+        """anonymous can't patch user to add or remove the group moderator."""
```
sampaccoud, line 315: not patching user anymore
```suggestion
        """anonymous user is not allowed to add or remove a user's moderator group via the action endpoint."""
```

tests/ashley/api/test_api_manage_moderators.py
```diff
@@ -211,13 +212,184 @@ def test_access_can_manage_moderators_list_non_moderators(self):
...
+        assign_perm("can_manage_moderator", user, forum, True)
+
+        # Creates the session
+        self.client.force_login(user, "ashley.auth.backend.LTIBackend")
+        session = self.client.session
+        session[SESSION_LTI_CONTEXT_ID] = lti_context.id
+        session.save()
+
+        action = random.choice(["add_group_moderator", "remove_group_moderator"])
+        response = self.client.post(f"/api/v1.0/users/{user.id}/{action}/")
```
sampaccoud, line 363: Note that POST should be on the list endpoint. That's why you get a 405. 
But ok you may want to confirm that this endpoint does not accept POST requests.

tests/ashley/api/test_api_manage_moderators.py
```diff
@@ -211,13 +212,184 @@ def test_access_can_manage_moderators_list_non_moderators(self):
...
+
+    def test_access_api_moderator_group_instructor_put(self):
+        """Instructor can't update user with a put to add or remove the group moderator."""
+        user = UserFactory(public_username="user")
+        # Creates a forum
+        lti_context = LTIContextFactory(lti_consumer=user.lti_consumer)
+        forum = ForumFactory()
+        forum.lti_contexts.add(lti_context)
+        assign_perm("can_manage_moderator", user, forum, True)
+        # Creates the session
```
sampaccoud, line 375:
```suggestion

        # Creates the session
```

tests/ashley/api/test_api_manage_moderators.py
```diff
@@ -211,13 +212,184 @@ def test_access_can_manage_moderators_list_non_moderators(self):
...
+        session.save()
+
+        action = random.choice(["add_group_moderator", "remove_group_moderator"])
+        response = self.client.put(f"/api/v1.0/users/{user.id}/{action}/")
+        self.assertEqual(response.status_code, 405)
+        self.assertEqual(response.json(), {"detail": 'Method "PUT" not allowed.'})
+
+    def test_access_api_can_manage_moderators_update_student_promote_revoke(self):
         """
         Promote and revoke a user with right context, permission, group
```
sampaccoud, line 388: not revoke, only adding the moderator group in this test.
Add another test for remove?

src/ashley/api/views.py
```diff
@@ -60,38 +60,43 @@ def get_queryset(self):
...
-            user.groups.add(group_moderator)
-            user.save()
-        else:
-            logger.debug(
-                "Update moderator forbidden - Role asked %s - user groups %s",
-                roles,
-                user_groups,
-            )
-            raise PermissionDenied()
+        if _FORUM_ROLE_MODERATOR not in user_groups and len(user_groups) > 0:
```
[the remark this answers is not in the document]
carofun: it's a security, a moderator can't be someone that is not in the original list of the user that can be a moderator.
we don't want to be able to add an unknown user, he must be related to this forum and be part of a group created for this forum. We could be even more restrictive in our test and make sure he's already a student. would you prefer this ?

src/frontend/js/data/fetchApi.ts
carofun: yes better

src/ashley/api/views.py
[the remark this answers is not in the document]
sampaccoud: ok... Then you should check that the group created for this forum is in user_groups! Checking that user_group > 0 is suboptimal and confusing if this is your intent?

src/ashley/api/views.py
```diff
@@ -92,8 +92,8 @@ def add_group_moderator(
         # Load current groups
         user_groups = lti_context.get_user_roles(user)
         # Asked to become moderator, check user is not yet moderator and make sure user
-        # belongs to this context by checking he has groups from this context.
-        if _FORUM_ROLE_MODERATOR not in user_groups and len(user_groups) > 0:
+        # belongs to this context by checking he has the group student from this context.
+        if _FORUM_ROLE_MODERATOR not in user_groups and _FORUM_ROLE_STUDENT in user_groups:
```
sampaccoud, line 96: Do you have a test that tries to go around the limitation and proves that it fails?

sampaccoud (dismissed):

sampaccoud (approved):

sampaccoud (approved):

PR #250, opened 2022-03-16
✨(forum) search links target directly the refered post

carofun:
## Purpose

A topic can have multiple posts and a lot of pages. Search results were targeting the topic on page 1 for any match.
If the post matching the search was on page 10, it made it hard to relate the search and the post. In this condition,
the results didn't seem accurate. To be more precise, we now build a link that directly targets the post on the right
page in the topic.

## Proposal

- build a direct link targeting the post

### DIFF
--- a/tests/ashley/test_forum_search.py
+++ b/tests/ashley/test_forum_search.py
@@ -158,7 +158,7 @@ class ForumSearchTestCase(TestCase):
         self.assertEqual(topic.subject, "5f3gh8ka1")
 
         # When we create the first post for this topic...
-        PostFactory(subject="497jk1sav", topic=topic)
+        post = PostFactory(subject="497jk1sav", topic=topic)
         # The topic subject should be overriden
         topic.refresh_from_db()
         self.assertEqual(topic.subject, "497jk1sav")
@@ -187,7 +187,8 @@ class ForumSearchTestCase(TestCase):
             response,
             (
                 f'<a href="/forum/forum/{forum.slug:s}-{forum.id:d}/topic/{topic.slug:s}-'
-                f'{topic.id:d}/" class="topic-name-link">{topic.subject:s}</a>'
+                f'{topic.id:d}/?post={ post.pk }#{ post.pk }" '
+                f'class="topic-name-link">{topic.subject:s}</a>'
             ),
             html=True,
         )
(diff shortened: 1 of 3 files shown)


tests/ashley/test_forum_search.py
```diff
@@ -730,3 +731,80 @@ def test_forum_search_paginated_links(self):
...
+        the topic. Using this link, the user get directed at the right
+        page of the topic the post is related to.
+        """
+
+        user = UserFactory()
+        post = PostFactory(text="azatofind")
+        PostFactory.create_batch(49, topic=post.topic)
+        # Create a specific post between two batchs
+        post_fourth_page = PostFactory(text="azatofind", topic=post.topic)
+        PostFactory.create_batch(50, topic=post.topic)
```
sampaccoud, line 747: this will slow down your tests considerably. You could avoid that if you could temporarily set pagination to 2 items per page for example?

tests/ashley/test_forum_search.py
```diff
@@ -730,3 +731,80 @@ def test_forum_search_paginated_links(self):
...
+        self.assertContains(response, direct_link_post_last_page)
+
+        # confirm the last page of the topic is page 7
+        response = self.client.get(url_topic)
+        self.assertContains(response, '<a href="?page=7" class="page-link">7</a>')
+        self.assertNotContains(response, '<a href="?page=8" class="page-link">8</a>')
+
+        response = self.client.get(direct_link_initial_post)
+        # first post is at page 1
+        self.assertContains(
```
sampaccoud, line 785: Not sure I understood correctly but calling the view with a "post" parameter in the querystring will redirect to the correct page at which the post is? If yes can you make it more explicit in the docstring by using the term "redirect"?

Using the Django redirect test would also make it clearer: https://docs.djangoproject.com/en/3.0/topics/testing/tools/#django.test.SimpleTestCase.assertRedirects

carofun: no it doesn't get redirected, it just loads the page directly at the right pagination. I've rephrased it to be more accurate.

sampaccoud: oh :astonished:  :+1:

sampaccoud (approved):

PR #259, opened 2022-05-16
🌐(i18n) generate i18n files

carofun:
## Purpose

Recent features added new strings to translate. We update the
repository by adding french translations.

## Proposal

Update translations in crowdin


sampaccoud:
@carofun it's ok if you add the fix release commit to this PR :wink:

src/ashley/locale/fr_FR/LC_MESSAGES/django.po
```diff
@@ -1085,12 +1086,16 @@ msgid "Can archive a forum"
 msgstr "Peut archiver un forum"
 
 #: machina_extensions/forum_permission/defaults.py:27
+msgid "Can manage a user's moderator status"
+msgstr "Permet de gérer le statut de modérateur d'un utilisateur"
```
sampaccoud, line 1090: I checked in Django and confirmed that permission names were not translated. After digging a little my understanding is that this name is destined to go to a model field that is not translatable anyway... https://github.com/django/django/blob/main/django/contrib/auth/models.py#L62

src/ashley/locale/fr_FR/LC_MESSAGES/django.po
```diff
@@ -1085,12 +1086,16 @@ msgid "Can archive a forum"
...
+#: machina_extensions/forum_permission/defaults.py:32
 msgid "Can rename a forum"
 msgstr "Peut renommer un forum"
 
-#: machina_extensions/forum_permission/defaults.py:32
-msgid "Can manage a user's moderator status"
-msgstr "Permet de gérer le statut de modérateur d'un utilisateur"
+#: machina_extensions/forum_permission/defaults.py:37
+msgid "Can lock forums of the course"
+msgstr "Peut verrouiller les forums du cours"
```
sampaccoud, line 1098: ditto

sampaccoud (approved):

PR #267, opened 2022-09-26
✨(forum) unlock all forums of an entire course

carofun:
## Purpose

A recent feature enabled administrators to lock the forums of a course. Use cases have shown that sometimes, the forums of the course need to be unlocked. This is the case, for instance, when a course is duplicated, the forums get copied but they keep the lock as well. It will now be possible to unlock forums once they have been locked.

## Proposal

- add new permission and enable to lock or unlock forums

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Allow unlocking all forums of a course
+
 ## [1.2.3] - 2022-06-13
 
 ### Changed
--- a/src/ashley/machina_extensions/forum/views.py
+++ b/src/ashley/machina_extensions/forum/views.py
@@ -12,7 +12,7 @@ from django.utils.http import urlencode
 from django.views.generic import UpdateView
 from machina.apps.forum.views import ForumView as BaseForumView
 from machina.apps.forum.views import IndexView as BaseIndexView
-from machina.apps.forum_permission.shortcuts import remove_perm
+from machina.apps.forum_permission.shortcuts import assign_perm, remove_perm
 from machina.apps.forum_permission.viewmixins import (
     PermissionRequiredMixin as BasePermissionRequiredMixin,
 )
@@ -249,12 +249,11 @@ class ForumArchiveView(PermissionRequiredMixin, UpdateView):
         return kwargs
 
 
-class ForumLockCourseView(PermissionRequiredMixin, UpdateView):
-    """View to lock forums of the course."""
+class ForumLockUnlockCourseView(PermissionRequiredMixin, UpdateView):
+    """Parent class to lock or unlock forums of the course."""
 
     model = Forum
     fields = ["lti_contexts"]
-    template_name = "forum/forum_lock.html"
 
     def get_forums_list(self):
         """Returns the list of forums of this course"""
@@ -274,6 +273,12 @@ class ForumLockCourseView(PermissionRequiredMixin, UpdateView):
         """Returns the success URL to redirect the user to."""
         return reverse("forum:index")
 
+
+class ForumLockCourseView(ForumLockUnlockCourseView):
+    """View to lock forums of the course."""
+
+    template_name = "forum/forum_lock.html"
+
     # pylint: disable=unused-argument
     def perform_permissions_check(self, user, obj, perms):
         """
@@ -289,7 +294,7 @@ class ForumLockCourseView(PermissionRequiredMixin, UpdateView):
     def post(self, request, *args, **kwargs):
         """
         We lock all forums of this course. Default group shouldn't have writing
-        permissions anymore. The LTIContex will then be marked as blocked.
+        permissions anymore. The LTIContext will then be marked as blocked.
         """
 
         lti_context = self.get_object().lti_contexts.get(
@@ -331,3 +336,45 @@ class ForumRenameView(PermissionRequiredMixin, UpdateView):
     def perform_permissions_check(self, user, obj, perms):
         """Performs the permissions check."""
         return self.request.forum_permission_handler.can_rename_forum(obj, user)
+
+
+class ForumUnlockCourseView(ForumLockUnlockCourseView):
+    """View to unlock forums of the course."""
+
+    template_name = "forum/forum_unlock.html"
+
+    # pylint: disable=unused-argument
+    def perform_permissions_check(self, user, obj, perms):
+        """
+        Performs the permissions check. If forums is not part of current LTIContext
+        access is refused.
+        """
+        if not obj.lti_contexts.filter(
+            id=self.request.forum_permission_handler.current_lti_context_id,
+            is_marked_locked=True,
+        ).exists():
+            return False
+        return self.request.forum_permission_handler.can_unlock_course(obj, user)
+
+    def post(self, request, *args, **kwargs):
+        """
+        We unlock all forums of this course. Default group should have writing
+        permissions. The LTIContext will then be marked as not blocked.
+        """
+
+        lti_context = self.get_object().lti_contexts.get(
+            id=self.request.forum_permission_handler.current_lti_context_id
+        )
+        default_group = lti_context.get_base_group()
+
+        # add all permissions for each forum of this LTIContext
+        for forum in self.get_forums_list():
+            # pylint: disable=not-an-iterable
+            for perm in DEFAULT_FORUM_BASE_WRITE_PERMISSIONS:
+                assign_perm(perm, default_group, forum, True)
+
+        # mark this course as unlocked
+        lti_context.is_marked_locked = False
+        lti_context.save()
+
+        return HttpResponseRedirect(self.get_success_url())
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Enables to unlock all forums of a course
```
sampaccoud, line 13:
```suggestion
- Allow unlocking all forums of a course
```

src/ashley/machina_extensions/forum/views.py
```diff
@@ -331,3 +336,45 @@ def get_success_url(self):
...
+            id=self.request.forum_permission_handler.current_lti_context_id,
+            is_marked_locked=True,
+        ).exists():
+            return False
+        return self.request.forum_permission_handler.can_unlock_course(obj, user)
+
+    def post(self, request, *args, **kwargs):
+        """
+        We unlock all forums of this course. Default group should have writing
+        permissions. The LTIContex will then be marked as not blocked.
```
sampaccoud, line 362:
```suggestion
        permissions. The LTIContext will then be marked as not blocked.
```

tests/ashley/test_course_unlocked.py
```diff
@@ -0,0 +1,473 @@
...
+    def _connects(self, role, forum_uuid=None, uuid=None, lis_person_sourcedid=None):
+        """
+        Utils not to repeat the connection of an instructor or
+        a student
+        """
+        consumer = LTIConsumerFactory(slug="consumer")
+        passport = LTIPassportFactory(title="consumer1_passport1", consumer=consumer)
+
+        if not forum_uuid:
+            forum_uuid = self.forum_uuid
```
sampaccoud, line 41:
```suggestion
        forum_uuid = forum_uuid or self.forum_uuid
```

tests/ashley/test_course_unlocked.py
```diff
@@ -0,0 +1,473 @@
...
+
+        # Build the LTI launch request
+        lti_parameters = {
+            "user_id": "643f1625-f240-4a5a-b6eb-89b317807963" if not uuid else uuid,
```
sampaccoud, line 45:
```suggestion
            "user_id": uuid,
```

tests/ashley/test_course_unlocked.py
```diff
@@ -0,0 +1,473 @@
...
+GroupForumPermission = get_model("forum_permission", "GroupForumPermission")
+
+
+class CourseUnlockCase(TestCase):
+    """Test the CourseUnlockCase class"""
+
+    forum_uuid = "8bb319aa-f3cf-4509-952c-c4bd0fb42fd7"
+    context_id = "course-v1:testschool+login+0001"
+
+    def _connects(self, role, forum_uuid=None, uuid=None, lis_person_sourcedid=None):
```
sampaccoud, line 32:
```suggestion
    def _connects(self, role, forum_uuid=None, uuid="643f1625-f240-4a5a-b6eb-89b317807963", lis_person_sourcedid="testuser"):
```

tests/ashley/test_course_unlocked.py
```diff
@@ -0,0 +1,473 @@
...
+        lti_parameters = {
+            "user_id": "643f1625-f240-4a5a-b6eb-89b317807963" if not uuid else uuid,
+            "lti_message_type": "basic-lti-launch-request",
+            "lti_version": "LTI-1p0",
+            "resource_link_id": "aaa",
+            "context_id": self.context_id,
+            "lis_person_contact_email_primary": "ashley@example.com",
+            "lis_person_sourcedid": "testuser"
+            if not lis_person_sourcedid
+            else lis_person_sourcedid,
```
sampaccoud, line 53:
```suggestion
            "lis_person_sourcedid": lis_person_sourcedid,
```

tests/ashley/test_course_unlocked.py
```diff
@@ -0,0 +1,473 @@
...
+
+        # permissions of instructor haven't changed
+        self.assertEqual(
+            count_perm_instructor,
+            GroupForumPermission.objects.filter(
+                forum=forum, group=instructor_group, has_perm=True
+            ).count(),
+        )
+
+        # for base group, original permissions have been reseted
```
sampaccoud, line 231:
```suggestion
        # for base group, original permissions have been reset
```

src/ashley/views.py
```diff
@@ -69,16 +69,25 @@ def _do_on_login(self, lti_request: LTI) -> HttpResponse:
                 type=Forum.FORUM_POST,
                 lti_contexts__id=context.id,
             )
-            # If course is locked, security to remove forum permission on default group
-            # only needed if the course has been marked outside web navigation
-            if context.is_marked_locked and any(
-                item in DEFAULT_FORUM_BASE_READ_PERMISSIONS
+            has_perm = any(
```
sampaccoud, line 72: took me too much time to understand and I think it would have helped if this variable had been called:
```suggestion
            has_any_write_permission = any(
```

Then now that I understand I think this value is correct to check if we are out of sync when the context is marked locked. But what we need to check if the context is marked unlocked is:
```python
 all(item in DEFAULT_FORUM_BASE_WRITE_PERMISSIONS for ...) 
```
right?

src/ashley/views.py
```diff
@@ -118,6 +112,44 @@ def _do_on_login(self, lti_request: LTI) -> HttpResponse:
...
+        """
+        The context contains a field is_marked_locked, we control that if
+        this field is checked or unchecked, there is no problem of unsynchronicity
+        It's a security to remove/add forum permission on the default group.
+        This could only happen if the course was locked/unlocked outside
+        web navigation directly on the database
+        """
+        # forum is marked locked but the base group still has all the writing
+        # permissions
+        locked_unsync = context.is_marked_locked and all(
```
sampaccoud, line 126: From what I understand, I don't think you're checking the right thing. Do you agree with the following statements or am I missing something?
- if if is marked locked then a synchronization issue would be that ANY write permission exists
- If it is marked unlocked, then a synchronization issue would be that NOT ALL write permissions are given

sampaccoud: If I'm right then there must also be missing tests to make sure it works correctly:
- To prove that the synchronization check for a locked forum works: create a locked forum and add 1 random write permission > assert that the check fails
- To prove that the synchronization check for an unlocked forum works: create an unlocked forum, add all write permissions except 1 taken randomly  > assert that the check fails

carofun: I've changed the behaviour regarding your last feedback. I'm going to change it back the way I've submitted it first to go back on my first idea.
I find it dangerous to consider a synchronization issue when 
* If it is marked unlocked, then a synchronization issue would be that NOT ALL write permissions are given
Permissions in some how could be changed outside this context, what is sure is that if none of the writing permissions are assigned whereas it's unlocked, it has to be a synchronization issue.

The same way if any write permissions exists I agee and it was this way first, there we could consider a synchronization issue.

sampaccoud: There's something I don't get: who assigns these write permissions? My understand is that it's done programmatically and that it's all or nothing. Is it not the case?

carofun: yes but it could be changed in the administration

src/ashley/views.py
```diff
@@ -118,6 +112,37 @@ def _do_on_login(self, lti_request: LTI) -> HttpResponse:
...
+            item in group_permissions for item in DEFAULT_FORUM_BASE_WRITE_PERMISSIONS
+        )
+
+        if locked_unsync or unlocked_unsync:
+            for perm in DEFAULT_FORUM_BASE_WRITE_PERMISSIONS:
+                # pylint: disable=expression-not-assigned
+                if locked_unsync:
+                    remove_perm(perm, context.get_base_group(), forum)
+                elif perm not in group_permissions:
+                    assign_perm(perm, context.get_base_group(), forum)
```
sampaccoud, line 144: Maybe clearer like this?
```suggestion
        if context.is_marked_locked:
            for item in group_permissions:
                if item in DEFAULT_FORUM_BASE_WRITE_PERMISSIONS:
                    # forum is marked locked but the base group still has a writing permission
                     remove_perm(perm, context.get_base_group(), forum)
        else:
            for item in DEFAULT_FORUM_BASE_WRITE_PERMISSIONS:
                if item not in group_permissions:
                    # forum is marked unlocked but the base group doesn't have all the writing permissions
                    assign_perm(perm, context.get_base_group(), forum)
```

tests/ashley/test_course_unlocked.py
```diff
@@ -0,0 +1,530 @@
...
+                GroupForumPermission.objects.filter(
+                    forum=forum, group=context.get_base_group(), has_perm=True
+                ).values_list("permission__codename", flat=True)
+            ),
+        )
+
+    def test_is_marked_unlocked_and_some_writing_perm(self):
+        """
+        Check when the tag is_marked_locked is set to False and that
+        there is still one writing permission for the base group,
```
sampaccoud, line 482:
```suggestion
        there are some missing writing permissions for the base group,
```

sampaccoud (approved):
Some minor comments then GTM :tada:

PR #268, opened 2022-10-05
🌐(i18n) generate i18n files

carofun:
## Purpose

Unlock feature added new strings to translate. We update the repository by adding french translations.

## Proposal

Description...

- add translations


src/ashley/locale/fr_FR/LC_MESSAGES/django.po
```diff
@@ -1087,15 +1087,19 @@ msgstr "Peut archiver un forum"
 
 #: machina_extensions/forum_permission/defaults.py:27
 msgid "Can manage a user's moderator status"
-msgstr ""
+msgstr "Permet de gérer le statut de modérateur d'un utilisateur"
```
sampaccoud, line 1090: To be coherent with other permissions it should be "Peut gérer..." ?

sampaccoud (approved):

PR #283, opened 2023-01-12
✨(front)add buttons in the toolbar of ashley for latex

carofun:
## Purpose

At the moment we can use mathematical latex syntax by pressing $ to start Inline Tex or  CMD + M for Tex Block.
For users, it would be much easier, if they have access directly in the toolbar to these commands. We, now, add two buttons to suggest these new features.

## Proposal

- add two buttons in the toolbar

https://user-images.githubusercontent.com/76939263/212353595-0d170700-63c7-40c4-b537-60be445418eb.mov

### update with new buttons

Capture d’écran 2023-01-18 à 10 17 43
Capture d’écran 2023-01-18 à 10 17 38
Capture d’écran 2023-01-18 à 10 17 34


src/frontend/scss/objects/_editor.scss
```diff
@@ -246,3 +246,16 @@ addImage Editor
     transform: scale(1);
   }
 }
+.latex-inline,
+.latex-inline:hover {
```
jbpenrath, line 250: Do you need the `:hover` selector to overwrite hover behavior ?
From user point of view, this is weird to remove hover effect. Here I would have rather add an effect on hover. (e.g: modify opacity ?)

src/frontend/scss/objects/_editor.scss
```diff
@@ -246,3 +246,16 @@ addImage Editor
...
+  display: inline-block;
+  background: url(../img/latex-inline.svg) 0 0 no-repeat;
+  background-size: 80%;
+}
+
+.latex-block,
+.latex-block:hover {
```
jbpenrath, line 257: ditto

src/frontend/js/components/AshleyEditor/LatexButton/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  theme: DraftJsButtonTheme;
+  getEditorState: () => EditorState;
+  setEditorState: (editorState: EditorState) => void;
+}
+
+export const LatexButton = (props: LatexButtonProps) => {
+  const node = useRef<HTMLDivElement>(null);
+  const { buttonWrapper = '', button } = props.theme;
+  const imageClassName =
+    props.type === TypeLatexStyle.INLINE ? 'latex-inline' : 'latex-block';
```
jbpenrath, line 20: Easier to maintain ?
```suggestion
  const imageClassName = useMemo(() => `latex-${props.type}`, [props.type]);
```

src/frontend/js/components/AshleyEditor/LatexButton/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+    e.preventDefault();
+    let newEditorState;
+    const editorState = props.getEditorState();
+    if (type === TypeLatexStyle.BLOCK) {
+      newEditorState = insertTeXBlock(editorState);
+    } else {
+      newEditorState = insertInlineTeX(editorState);
+    }
+    props.setEditorState(newEditorState);
+    return 'handled';
```
jbpenrath, line 34: Is `"handled"` something waited by draft.js to know to know which type of action has been triggered ? If yes, what do you think to create a enum to store those action names ?

src/frontend/js/components/AshleyEditor/LatexButton/index.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+}));
+
+describe('LatexButton', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+  beforeEach(() => {
+    props.editorState = EditorState.createEmpty();
+    props.getEditorState = jest.fn().mockReturnValue(props.editorState);
+  });
```
jbpenrath, line 30: Do you really need to reset those values ?

jbpenrath (commented):
Wait for better icons

src/frontend/scss/objects/_editor.scss
carofun: Yes, I add to overwrite because of draft.js otherwise I had a gray square instead. True, and good idea I can still implement an opacity property on the over ver

src/frontend/js/components/AshleyEditor/LatexButton/index.spec.tsx
carofun: at one point yes, but not after my resetMock and I didn't erased it, good catch

src/frontend/i18n/locales/fr_FR.json
```diff
@@ -15,6 +15,14 @@
     "description": "Error message on uploading image, max size reached",
     "message": "Une erreur est survenue lors du téléchargement de l'image, la taille du fichier est supérieure à {maxSize}Mo."
   },
+  "components.AshleyEditor.LatexButton.toolTipAddLatexBlock": {
+    "description": "Message over the button to add a Latex block formula",
+    "message": "Ajouter une formule mathématique Latex dans un nouveau bloc"
+  },
+  "components.AshleyEditor.LatexButton.toolTipAddLatexInline": {
+    "description": "Message over the button to add a Latex inline formula",
+    "message": "Ajouter une formule mathématique en ligne Latex"
```
jbpenrath, line 24: What do you think to use less technical sentence ?
```suggestion
    "message": "Insérer une formule mathématique Latex à l'emplacement du curseur."
```

src/frontend/js/components/AshleyEditor/LatexButton/index.tsx
```diff
@@ -9,17 +9,62 @@ import { FormattedMessage } from 'react-intl';
...
+      ? selection.getStartOffset()
+      : selection.getEndOffset();
+    const entityKey = block.getEntityAt(offset);
+    if (
+      entityKey &&
+      contentState.getEntity(entityKey).getType() === 'INLINETEX'
+    ) {
+      return true;
+    }
+    return false;
```
jbpenrath, line 44: Could be simplified : 

```suggestion
    const selection = props.editorState.getSelection();
    const contentState = props.editorState.getCurrentContent();
    const blockKey = selection.getStartKey();
    const block = contentState.getBlockForKey(blockKey);
    const offset = selection.getIsBackward()
          ? selection.getStartOffset()
          : selection.getEndOffset();
    const entityKey = block?.getEntityAt(offset);

    if (!entityKey) {
        return false;
    }

    return contentState.getEntity(entityKey).getType() === 'INLINETEX';
```

jbpenrath (approved):
LGTM ✅ 💪

PR #299, opened 2024-11-14
⬆️(compose) upgrade to postgres 13

lunika:
## Purpose

We want to use postgres 13, we upgrade the docker image use in the docker compose configuration.

## Proposal

- [x] upgrade to postgres 13

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,9 +172,9 @@ jobs:
           ELASTICSEARCH_PORT: 9200
           DJANGO_AWS_S3_REGION_NAME: eu-west-1
       # services
-      - image: circleci/postgres:12-ram
+      - image: cimg/postgres:16.4
         environment:
-          POSTGRES_DB: test_ashley
+          POSTGRES_DB: ashley
           POSTGRES_USER: fun
           POSTGRES_PASSWORD: pass
       - image: elasticsearch:5
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,8 +96,7 @@ jobs:
       # Generate a version.json file describing app release
       - <<: *generate-version-file
       # Activate docker-in-docker
-      - setup_remote_docker:
-          version: 19.03.13
+      - setup_remote_docker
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -15,13 +15,13 @@
...
 # -- Base image --
-FROM python:3.8-slim as base
+FROM python:3.8-slim AS base
 
 # Upgrade pip to its latest release to speed up dependencies installation
 RUN pip install --upgrade pip
 
 # ---- Front-end builder image ----
-FROM node:14 as front-builder
+FROM node:14 AS front-builder
```
claudusd, line 24: You Upgrade the node's image in the CI definition and the compose file. 
Maybe here too ?
```suggestion
FROM node:16 AS front-builder
```

wilbrdt (approved):

claudusd (approved):

kernicPanel (commented):
We still have `circleci/postgres:12-ram` in https://github.com/openfun/ashley/blob/030c10e58db5b0ec39bf05756061159b0dfd11bd/.circleci/config.yml#L174C16-L174C40

lunika:
> We still have `circleci/postgres:12-ram` in https://github.com/openfun/ashley/blob/030c10e58db5b0ec39bf05756061159b0dfd11bd/.circleci/config.yml#L174C16-L174C40

Nice catch

.circleci/config.yml
```diff
@@ -172,7 +171,7 @@ jobs:
           ELASTICSEARCH_PORT: 9200
           DJANGO_AWS_S3_REGION_NAME: eu-west-1
       # services
-      - image: circleci/postgres:12-ram
+      - image: circleci/postgres:17-ram
```
wilbrdt, line 174: This tag does not seem to exist on DockerHub
And if I remember correctly, the `circleci/` images are deprecated, should we take this opportunity to upgrade to `cimg/` images?

wilbrdt: There's no RAM variant anymore, but should pass with `cimg/postgres:17.0`

